PR #1, opened 2015-02-03
Restructre the apprtc repo

jiayliu:
Now GAE must be pushed from out/app_engine, created from 'grunt build'.


juberti:
Move third_party under /src, as in Chrome
Also, adapter.js seems out of sync with the latest in webrtc/adapter. Long term, we should make apprtc pull adapter as a dependency, but we can do that in a followup.

jiayliu:
Done. PTAL.

juberti:
lgtm. Please move over any remaining apprtc issues from the issue tracker and PSA this to discuss-webrtc.

PR #2, opened 2015-02-04
Fix the issue of inactive infobox making sharing link unclickable

jiayliu:
The infobox is rendered as opacity:0 when inactive, covering the sharing link and making it unclickable.
Fixed by setting display:none when inactive.


jiayliu:
@chuckhays

chuckhays:
LGTM.

PR #24, opened 2015-02-04
Fix build script on mac.

chuckhays:
The find/cp line used flags not present on mac. The single line command that works for mac (find -E $SRC_DIR/app_engine -regex ".*\.(py|yaml|html)" -not -name '*test.py*') doesn't work on linux.


jiayliu:
lgtm

Issue #21, opened 2015-02-04
The Analytics change breaks AppRTC test in WebRTC waterfall 


jiayliu:
https://github.com/GoogleChrome/webrtc/issues/467#event-230065986

phoglund commented 6 hours ago
I reverted the analytics change for now.
When bringing up the server on the bots after the analytics change:

INFO 2015-02-04 09:56:16,612 devappserver2.py:557] Skipping SDK update check.
WARNING 2015-02-04 09:56:16,622 simple_search_stub.py:1002] Could not read search indexes from /tmp/appengine.apprtc.chrome-bot/search_indexes
INFO 2015-02-04 09:56:16,623 api_server.py:138] Starting API server at: http://localhost:58086
INFO 2015-02-04 09:56:16,634 dispatcher.py:164] Starting module "default" running at: http://localhost:9999
ERROR 2015-02-04 09:56:17,886 wsgi.py:219]
Traceback (most recent call last):
File "/b/build/slave/Linux_Tester/build/google_appengine/google/appengine/runtime/wsgi.py", line 196, in Handle
handler = config_handle.add_wsgi_middleware(self.LoadHandler())
File "/b/build/slave/Linux_Tester/build/google_appengine/google/appengine/runtime/wsgi.py", line 255, in _LoadHandler
handler = __import(path[0])
File "/b/build/slave/Linux_Tester/build/src/out/webrtc-samples/samples/web/content/apprtc/apprtc.py", line 26, in 
import analytics
File "/b/build/slave/Linux_Tester/build/src/out/webrtc-samples/samples/web/content/apprtc/analytics.py", line 14, in 
from oauth2client.client import SignedJwtAssertionCredentials
ImportError: cannot import name SignedJwtAssertionCredentials
INFO 2015-02-04 09:56:17,891 module.py:593] default: "GET /r/some_room?wshpp=localhost:8089&wstls=false&firefox_fake_device=1 HTTP/1.1" 500 -
ERROR 2015-02-04 09:56:17,919 wsgi.py:219]
Traceback (most recent call last):
File "/b/build/slave/Linux_Tester/build/google_appengine/google/appengine/runtime/wsgi.py", line 196, in Handle
handler = config_handle.add_wsgi_middleware(self.LoadHandler())
File "/b/build/slave/Linux_Tester/build/google_appengine/google/appengine/runtime/wsgi.py", line 255, in _LoadHandler
handler = __import(path[0])
[...]

PR #26, opened 2015-02-04
Disable analytics for server exception on dev server

jiayliu:
@milkypostman


jiayliu:
https://github.com/webrtc/apprtc/issues/21

milkypostman:
lgtm

milkypostman:
You should probably remove the import just to be safe as well.

PR #33, opened 2015-02-04
Add a 2h expiration time for static files

jiayliu:
This is one of the issues found by PageSpeed Insights:

Leverage browser caching
Setting an expiry date or a maximum age in the HTTP headers for static resources instructs the browser to load previously downloaded resources from local disk rather than over the network.
Leverage browser caching for the following cacheable resources:
https://www.google-analytics.com/analytics.js (2 hours)

### DIFF
--- a/src/app_engine/app.yaml
+++ b/src/app_engine/app.yaml
@@ -3,6 +3,7 @@ version: 2
 runtime: python27
 threadsafe: true
 api_version: 1
+default_expiration: "2h"
 
 handlers:
 - url: /html


juberti:
Please explain the intent of this change in the PR

juberti:
Sorry, I still don't get it - how does making our static content less cacheable solve the analytics.js problem?

jiayliu:
Oops, I thought they are not cached, but it seems they currently have expiration time as 1y.

PR #35, opened 2015-02-04
Notify GCM endpoints on call start, accept and decline.

tkchin:
@jiayliu @chuckhays 

Tested using some local JS. Was able to receive GCM messages on two Chrome endpoints.

### DIFF
--- a/src/app_engine/decline_page.py
+++ b/src/app_engine/decline_page.py
@@ -2,7 +2,7 @@
 #
 # Copyright 2015 Google Inc. All Rights Reserved.
 
-"""AppRTC Decline Call Handler
+"""AppRTC Decline Call Handler.
 
 This module implements declining a call.
 """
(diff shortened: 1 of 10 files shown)


src/app_engine/decline_page.py
```diff
@@ -12,36 +12,99 @@
...
+
+    # Check that room exists.
+    if declined_room is None:
+      logging.warning(
+          'Client:%s declined unknown room:%s', callee_gcm_id, room_id)
+      self.write_response(constants.RESPONSE_INVALID_ROOM)
+      return
+
+    # Check that the room supports decline.
+    if declined_room.room_type != room.Room.TYPE_DIRECT:
```
jiayliu: All the checks should happen inside the cas loop to avoid races.
Why now do remove_room first, if it succeeds, then send GCM, so that you don't need to do checks outside remove_room.

PR #39, opened 2015-02-05
Rename /bind to /register

jiayliu:


jiayliu:
@chuckhays

src/app_engine/apprtc.py
```diff
@@ -122,11 +122,13 @@ def get(self):
 
 app = webapp2.WSGIApplication([
     ('/', MainPage),
+    # TODO(jiayl): Remove the deprecated API when Android is updated.
```
tkchin: nit: move the deprecated stuff to the bottom so it's clear which ones are to be removed

tkchin:
lgtm

PR #40, opened 2015-02-05
Send relevant GCM messages on call start / accept / decline.

tkchin:
@jiayliu @chuckhays

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -66,7 +66,13 @@ module.exports = function(grunt) {
         }
       },
       buildAppEnginePackage: {
-        command: './build/build_app_engine_package.sh',
+        command: function(apiKey) {
+          var cmd = './build/build_app_engine_package.sh';
+          if (apiKey) {
+            cmd += ' ' + apiKey;
+          }
+          return cmd;
+        },
         options: {
           stdout: true,
           stderr: true
@@ -173,7 +179,16 @@ module.exports = function(grunt) {
                                  'shell:buildVersion', 'runPythonTests', 'jstests']);
   grunt.registerTask('runPythonTests', ['shell:buildAppEnginePackage', 'shell:runPythonTests']);
   grunt.registerTask('jstests', ['closurecompiler:debug', 'jstdPhantom']);
-  grunt.registerTask('build', ['closurecompiler:debug', 'shell:buildVersion', 'shell:buildAppEnginePackage', 'grunt-chrome-build']);
+  grunt.registerTask('build', function(apiKey) {
+    var appEngineTask = 'shell:buildAppEnginePackage';
+    if (apiKey) {
+      appEngineTask += ':' + apiKey;
+    }
+    grunt.task.run(['closurecompiler:debug',
+                    'shell:buildVersion',
+                    appEngineTask,
+                    'grunt-chrome-build']);
+  });
   // also possible to call JavaScript directly in registerTask()
   // or to call external tasks with grunt.loadTasks()
 };
--- a/src/app_engine/decline_page.py
+++ b/src/app_engine/decline_page.py
@@ -2,7 +2,7 @@
 #
 # Copyright 2015 Google Inc. All Rights Reserved.
 
-"""AppRTC Decline Call Handler
+"""AppRTC Decline Call Handler.
 
 This module implements declining a call.
 """
@@ -12,36 +12,52 @@ import logging
 import webapp2
 
 import constants
-import gcmrecord
+import gcm_notify
 import room
 import util
+
+from gcmrecord import GCMRecord
 
 class DeclinePage(webapp2.RequestHandler):
   def write_response(self, result):
     self.response.write(json.dumps({
-      'result': result
+        'result': result
     }))
-  
+
   def post(self, room_id):
     msg = util.get_message_from_json(self.request.body)
     if not util.has_msg_field(msg, constants.PARAM_CALLEE_GCM_ID, basestring):
       self.write_response(constants.RESPONSE_INVALID_ARGUMENT)
       return
-
     callee_gcm_id = msg[constants.PARAM_CALLEE_GCM_ID]
 
-    # Look up and validate callee by gcm id.
-    # TODO (chuckhays): Once registration is enabled, turn on restriction
+    # Remove the room.
+    result, declined_room = room.remove_room_for_declined_call(
+        self.request.host_url, room_id, callee_gcm_id)
+    if result != constants.RESPONSE_SUCCESS:
+      self.write_response(result)
+      return
+
+    # We successfully removed the room from memcache, but should still have a
+    # copy of the removed room. Use that to get the caller's GCM id.
+    caller_gcm_id = declined_room.get_other_client_id(callee_gcm_id)
+
+    # Retrieve callee's GCM ids.
+    # TODO(chuckhays): Once registration is enabled, turn on restriction
     # to only return verified records.
-    callee_records = gcmrecord.GCMRecord.get_by_gcm_id(callee_gcm_id, False)
+    callee_records = GCMRecord.get_associated_records_for_gcm_id(
+        callee_gcm_id, False)
     if len(callee_records) < 1:
       self.write_response(constants.RESPONSE_INVALID_CALLEE)
       return
 
-    # TODO (chuckhays): Send message to ringing clients to stop ringing.
-    # TODO (chuckhays): Send message to caller that call was declined.
+    # Notify caller and other callee endpoints about the call decline.
+    gcm_ids_to_notify = [caller_gcm_id]
+    for record in callee_records:
+      if record.gcm_id != callee_gcm_id:
+        gcm_ids_to_notify.append(record.gcm_id)
+    gcm_notify.send_byes(
+        gcm_ids_to_notify, room_id, gcm_notify.GCM_MESSAGE_REASON_TYPE_DECLINED)
 
-    result = room.remove_room_for_declined_call(self.request.host_url,
-                                                room_id, callee_gcm_id)
     self.write_response(result)
 
--- a/src/app_engine/gcm_notify.py
+++ b/src/app_engine/gcm_notify.py
@@ -1,0 +1,136 @@
+#!/usr/bin/python2.7
+# Copyright 2015 Google Inc. All Rights Reserved.
+"""Module for communication with GCM server.
+"""
+
+import json
+import logging
+import os
+import time
+
+import constants
+
+from google.appengine.api import urlfetch
+
+# Load GCM API key from configuration file.
+GCM_API_KEY = None
+GCM_API_URL = 'https://android.googleapis.com/gcm/send'
+
+GCM_MESSAGE_TYPE_BYE = 'BYE'
+GCM_MESSAGE_TYPE_INVITE = 'INVITE'
+GCM_MESSAGE_TYPE_KEY = 'type'
+GCM_MESSAGE_REASON_KEY = 'reason'
+GCM_MESSAGE_REASON_TYPE_DECLINED = 'calleeDeclined'
+GCM_MESSAGE_REASON_TYPE_ACCEPTED = 'calleeAccepted'
+GCM_MESSAGE_REASON_TYPE_HANGUP = 'callerHangup'
+# Time that messages are kept on GCM server in seconds. Two days.
+GCM_TIME_TO_LIVE_IN_SECONDS = 2 * 24 * 60 * 60
+
+
+def get_gcm_api_key():
+  """Reads GCM API key from disk and caches it."""
+  global GCM_API_KEY
+  if GCM_API_KEY is not None:
+    return GCM_API_KEY
+  try:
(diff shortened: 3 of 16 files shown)


tkchin:
@jiayliu @chuckhays PTAL

src/app_engine/decline_page.py
```diff
@@ -12,36 +12,54 @@
...
+
+    # We successfully removed the room from memcache, but should still have a
+    # copy of the removed room. Use that to get the caller's GCM id.
+    caller_gcm_id = declined_room.get_other_client_id(callee_gcm_id)
+
+    # Retrieve callee's other GCM ids.
+    # TODO(chuckhays): Once registration is enabled, turn on restriction
     # to only return verified records.
-    callee_records = gcmrecord.GCMRecord.get_by_gcm_id(callee_gcm_id, False)
+    callee_id = gcmrecord.GCMRecord.get_user_id_for_gcm_id(callee_gcm_id, False)
```
jiayliu: nit: put get_user_id_for_gcm_id and get_by_user_id in a helper method of gcmrecord. We should need it in other places as well.

src/app_engine/gcm_config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "GCM_API_KEY": ""
```
jiayliu: Please explain how to populate this file in README

src/app_engine/gcm_notify.py
```diff
@@ -0,0 +1,108 @@
...
+    return
+
+  def handle_rpc_result(rpc):
+    result = rpc.get_result()
+    status_code = result.status_code
+    if status_code != 200:
+      logging.error('Failed to send GCM message. Result:%d', status_code)
+
+  headers = {
+      'Content-Type': 'application/json',
```
jiayliu: Should we set collapse_key and time_to_live?

src/app_engine/gcm_notify.py
```diff
@@ -0,0 +1,108 @@
...
+    logging.error('Failed to open GCM config file: %s', str(e))
+  return GCM_API_KEY
+
+
+def send_gcm_messages(gcm_ids, message):
+  """Sends the message to each endpoint specified in gcm_ids.
+
+  Makes asynchronous posts to the GCM server with the appropriate payload
+  format. GCM server will then deliver the message the each endpoint. Will
+  block until all posts complete or fail.
```
chuckhays: > Will block until all posts complete or fail.

Do we need to block? This will delay the response to the client right?

src/app_engine/room.py
```diff
@@ -103,42 +108,44 @@ def remove_room_for_declined_call(host, room_id, callee_gcm_id):
...
       logging.warning('Can\'t remove room ' + room_id +
           'because it is full (' + str(room.get_occupancy()) +') client: ' +
           callee_gcm_id)
-      return constants.RESPONSE_INVALID_ROOM
+      return (constants.RESPONSE_INVALID_ROOM, room)
 
     # Reset the room to the initial state so it may be reused.
-    room.reset()
-    if memcache_client.cas(key, room, constants.ROOM_MEMCACHE_EXPIRATION_SEC):
+    reset_room = Room(room.room_type)
```
chuckhays: We can probably remove the reset() method if we aren't using it here.

chuckhays:
I likely made a lot of the same lint fixes you did, hope that doesn't cause too much pain.

src/app_engine/join_page.py
```diff
@@ -65,37 +69,39 @@ def handle_call(self, msg, room_id):
...
-          ' callee id: ' + callee_id)
+                      'initiate a new call. Caller gcm id: ' + caller_gcm_id +
+                      ' callee id: ' + callee_id)
       self.report_error(constants.RESPONSE_INVALID_ROOM)
       return
 
-    allowed_gcm_ids = map(lambda x: x.gcm_id, callee_records)
-    allowed_gcm_ids.append(caller_gcm_id)
+    callee_gcm_ids = [record.gcm_id for record in callee_records]
+    allowed_gcm_ids = callee_gcm_ids + [caller_gcm_id]
```
jiayliu: Lint found this? Interesting.

src/app_engine/join_page.py
```diff
@@ -107,42 +113,53 @@ def handle_accept(self, msg, room_id):
...
+      logging.info('Error adding client to room: ' + result['error'] +
+                   ', room_state=' + result['room_state'])
       self.write_response(result['error'], {}, [])
       return
 
-    # TODO (chuckhays): Send message to ringing clients to stop ringing.
+    # Look up callee's other GCM endpoints.
+    # TODO(chuckhays): Once registration is enabled, turn on restriction
+    # to only return verified records.
+    callee_records = gcmrecord.GCMRecord.get_by_user_id(callee_id, False)
```
jiayliu: ditto helper method

src/app_engine/gcm_notify.py
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/python2.4
+# Copyright 2015 Google Inc. All Rights Reserved.
+"""Module for communication with GCM server.
```
jiayliu: unit test? We can stub out the gcm API in a similar way as analytics_test. Probably moving the mocking code test_unitilities.

src/app_engine/gcm_notify.py
tkchin: We don't have way to query server about who is calling a given device at the moment. So e.g. if device is offline and 5 people call, it gets collapsed into one message and the other devices end up timing out. I suppose this isn't unreasonable and it's better than having really old devices get inundated with messages it needs to handle on startup. It does limit our ability to have a 'missed calls' UI though. WDYT?

TTL makes sense.

jiayliu: How about using the room id as the collapse key?

On Thu, Feb 5, 2015 at 11:34 AM, Zeke Chin <notifications@github.com> wrote:

> In src/app_engine/gcm_notify.py
> <https://github.com/webrtc/apprtc/pull/40#discussion_r24191795>:
>
> > +    gcm_ids: List of verified GCM ids.
> > +    message: Dictionary containing custom data.
> > +  """
> > +  gcm_api_key = get_gcm_api_key()
> > +  if gcm_api_key is None:
> > +    logging.warning('Did not send GCM message due to missing API key.')
> > +    return
> > +
> > +  def handle_rpc_result(rpc):
> > +    result = rpc.get_result()
> > +    status_code = result.status_code
> > +    if status_code != 200:
> > +      logging.error('Failed to send GCM message. Result:%d', status_code)
> > +
> > +  headers = {
> > +      'Content-Type': 'application/json',
>
> We don't have way to query server about who is calling a given device at
> the moment. So e.g. if device is offline and 5 people call, it gets
> collapsed into one message and the other devices end up timing out. I
> suppose this isn't unreasonable and it's better than having really old
> devices get inundated with messages it needs to handle on startup. It does
> limit our ability to have a 'missed calls' UI though. WDYT?
>
> TTL makes sense.
>
> —
> <https://github.com/webrtc/apprtc/pull/40/files#r24191795>.
>



-- 
Jiayang Liu

src/app_engine/gcm_notify.py
tkchin: If we don't we won't know result of the fetch. We could spin up a thread that does all of this and return early.. However, the posts should be fast enough I think, since it should be bouncing around Google servers. The default is that rpc will timeout in 5s so that's the maximum delay. Added TODO for checking timing / performance.

jiayliu: Could you add logging messages at the beggining and the end of the method, so we can easily see the timing from the server log?

src/app_engine/gcm_notify.py
```diff
@@ -0,0 +1,108 @@
...
+  gcm_api_key = get_gcm_api_key()
+  if gcm_api_key is None:
+    logging.warning('Did not send GCM message due to missing API key.')
+    return
+
+  def handle_rpc_result(rpc):
+    result = rpc.get_result()
+    status_code = result.status_code
+    if status_code != 200:
+      logging.error('Failed to send GCM message. Result:%d', status_code)
```
jiayliu: Should we return this error to the user? Since we failed to ring the callee.

src/app_engine/gcm_notify.py
tkchin: Still same issue, just that there's a buffer. GCM only supports 4 collapse keys.

tkchin:
hitting pause button to wait for Chuck's change to go in so I can merge it and add the last GCM call. Will be writing unit tests in the meantime.

chuckhays:
Mine is merged now.

src/app_engine/gcm_notify.py
tkchin: I don't know that it's worth it to add another return code to check. Also, it would be unclear what we would say if we failed at one but succeeded at the rest if there are multiple devices. The scenario is similar to when user tries to ring a device that is currently offline. It should be pretty rare that we are unable to send request from one Google server to another. WDYT?

tkchin:
PTAL

src/app_engine/gcm_notify.py
jiayliu: OK, I think we can live with it and we'll have the ring-ack message to notify the caller.

src/app_engine/gcm_notify.py
```diff
@@ -0,0 +1,135 @@
...
+    rpcs.append(rpc)
+
+  # Wait for all the requests to finish.
+  # TODO(tkchin): check performance on live server and multiple devices. Move
+  # requests to separate thread and return early if needed.
+  rpc_wait_start_time = time.time()
+  for rpc in rpcs:
+    rpc.wait()
+  rpc_wait_end_time = time.time()
+  logging.info('Took %.3fs to send GCM messages, spent %.3fs waiting for rpcs '
```
jiayliu: nit: log the # of messages sent?

src/app_engine/gcm_config_DO_NOT_CHECK_IN.json
```diff
@@ -0,0 +1,3 @@
+{
+  "GCM_API_KEY": ""
```
jiayliu: can we just git ignore it?

tkchin: can't git ignore a file that's already tracked (ie committed). It'll still show up if you edit it.

jiayliu: What about not checking in this file? Just show an example of the file in
README.
We could also modify the build script to generate the temple.

On Fri, Feb 6, 2015 at 11:50 AM, Zeke Chin <notifications@github.com> wrote:

> In src/app_engine/gcm_config_DO_NOT_CHECK_IN.json
> <https://github.com/webrtc/apprtc/pull/40#discussion_r24266785>:
>
> > @@ -0,0 +1,3 @@
> > +{
> > +  "GCM_API_KEY": ""
>
> can't git ignore a file that's already tracked (ie committed). It'll still
> show up if you edit it.
>
> —
> <https://github.com/webrtc/apprtc/pull/40/files#r24266785>.
>



-- 
Jiayang Liu

tkchin: I'll modify build_app_engine_package.sh to take an optional argument then.

tkchin: Added scripts for generating json file. PTAL.

Gruntfile.js
```diff
@@ -173,7 +179,16 @@ module.exports = function(grunt) {
...
+  grunt.registerTask('build', function(apiKey) {
+    var appEngineTask = 'shell:buildAppEnginePackage';
+    if (apiKey) {
+      appEngineTask += ':' + apiKey;
+    }
+    grunt.task.run(['closurecompiler:debug',
+                    'shell:buildVersion',
+                    appEngineTask,
+                    'grunt-chrome-build']);
+  });
```
jiayliu: Is the syntax to call  "grunt build:API_KEY"?
Please clarify in ReadMe.

tkchin: Yup.
Oops, forgot to update that. Updated.

jiayliu:
.gitignore is still not in the change list.

tkchin:
But it shouldn't be in the changelist. Everything in out/ is already git ignored.
The previous update was to remove it from the CL (for when I found out that git ignoring tracked files doesn't do anything).
Am I missing something?

jiayliu:
Haha, you are right. lgtm

On Fri, Feb 6, 2015 at 2:44 PM, Zeke Chin <notifications@github.com> wrote:

> But it shouldn't be in the changelist. Everything in out/ is already git
> ignored.
> The previous update was to remove it from the CL (for when I found out
> that git ignoring tracked files doesn't do anything).
> Am I missing something?
>
> —
> <https://github.com/webrtc/apprtc/pull/40#issuecomment-73327107>.
>



-- 
Jiayang Liu

chuckhays:
LGTM as well.

Issue #4, opened 2015-02-04
Toolbar buttons don't show after rejoining room


jiayliu:
Ported from https://github.com/GoogleChrome/webrtc/issues/457

If you leave and then rejoin a room via the 'rejoin' button, the toolbar buttons never show up.

chuckhays commented 6 hours ago
This was the result merging two features working in the same area. PR #432 refactored code in a way that no longer worked in the way assumed by #424.

PR #45, opened 2015-02-05
Fix for #4, rejoin is broken

chuckhays:
This is the same code that was in https://github.com/webrtc/samples/pull/462, but moved to the new repo and added a unit test.

@jiayliu @juberti

### DIFF
--- a/src/web_app/js/call.js
+++ b/src/web_app/js/call.js
@@ -42,10 +42,10 @@ var Call = function(params) {
 
   this.getMediaPromise_ = null;
   this.getTurnServersPromise_ = null;
-  this.initPromises_();
+  this.requestMediaAndTurnServers_();
 };
 
-Call.prototype.initPromises_ = function() {
+Call.prototype.requestMediaAndTurnServers_ = function() {
   this.getMediaPromise_ = this.maybeGetMedia_();
   this.getTurnServersPromise_ = this.maybeGetTurnServers_();
 };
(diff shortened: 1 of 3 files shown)


src/web_app/js/call_test.js
```diff
@@ -0,0 +1,58 @@
...
+/* globals TestCase, SignalingChannel:true, requestUserMedia:true,
+   assertEquals, assertTrue */
+
+'use strict';
+
+var MockSignalingChannel = function() {
+};
+
+MockSignalingChannel.prototype.open = function() {
+  return new Promise(function(resolve) { resolve(); });
```
jiayliu: nit: just return Promise.resolve()?

src/web_app/js/call_test.js
```diff
@@ -0,0 +1,58 @@
...
+  SignalingChannel = this.signalingChannelBackup_;
+  requestUserMedia = this.requestUserMediaBackup_;
+};
+
+CallTest.prototype.testRestartInitializesMedia = function() {
+  var call = new Call(this.params_);
+  var mediaStarted = false;
+  call.onlocalstreamadded = function(stream) {
+    mediaStarted = true;
+    assertEquals('stream', stream);
```
jiayliu: nit: move 'stream' to a const var

src/web_app/js/call.js
```diff
@@ -40,6 +40,12 @@ var Call = function(params) {
   this.onsignalingstatechange = null;
   this.onstatusmessage = null;
 
+  this.getMediaPromise_ = null;
+  this.getTurnServersPromise_ = null;
+  this.initPromises_();
```
jiayliu: nit: a more descriptive name than initPromise_, e.g. requestMediaAndTurnServers.

src/web_app/js/call_test.js
chuckhays: Sure, I had that originally, I think I changed it while trying to debug the Promise.all issue.

chuckhays:
Updated. PTAL.

jiayliu:
lgtm

PR #46, opened 2015-02-05
Fix broken tests, rename constants.

chuckhays:
As per our previous discussion, this supports having client_id on the body as well as the original method for existing clients.

@jiayliu @tkchin


src/app_engine/message_page.py
```diff
@@ -43,10 +44,19 @@ def send_message_to_collider(self, room_id, client_id, message):
...
 
-  def post(self, room_id, client_id):
-    message_json = self.request.body
+  def post(self, room_id, client_id = None):
+    message = self.request.body
+    if client_id is None:
+      # Check for client_id in body.
+      msg = util.get_message_from_json(self.request.body)
+      if not util.has_msg_fields(msg, ((constants.PARAM_USER_GCM_ID, basestring), (constants.PARAM_MESSAGE, basestring))):
+        self.write_response(constants.RESPONSE_INVALID_ARGUMENT)
```
jiayliu: Have a test for this?

src/app_engine/message_page.py
```diff
@@ -43,10 +44,19 @@ def send_message_to_collider(self, room_id, client_id, message):
...
+        return
+      client_id = msg[constants.PARAM_USER_GCM_ID]
+      message = msg[constants.PARAM_MESSAGE]
+
     result = room_module.save_message_from_client(
```
jiayliu: Not introduced in this CL, but we have UNKNOWN_ROOM/UNKNOWN_CLIENT, duplicated with INVALID_ROOM/INVALID_USER. Maybe this is a good chance to unify them?

src/app_engine/message_page.py
chuckhays: Do you mean the INVALID_ARGUMENT case?

Probably a good idea.

jiayliu: Right, invalid_argument

chuckhays:
Updated with suggestions and fixed pylint issues. PTAL.

src/app_engine/test_utilities.py
```diff
@@ -42,13 +42,19 @@ def tearDown(self):
     self.testbed.deactivate()
     analytics.report_event = self.oldReportEvent
 
-  def checkInvalidRequests(self, path, params):
+  def checkInvalidRequests(self, path, params, jsonResult=False):
```
jiayliu: Which API returns this form of result? We probably should make it consistent with the other APIs.

chuckhays: I modeled the new endpoints and my updates on the join page, which returns json. You are right that things are not consistent. I think everything should return json with a result field. Otherwise clients would have to try to parse the result as json, if it fails, try to parse it as a result string. Seems like we can standardize on a message format.

tkchin: +1 to json with result field

chuckhays:
Anyone want to take a look at this?

jiayliu:
I think we should put a hold on this, and work on using a separate client
id instead, so that we don't need to expose gcmId to all the websocket
server APIs.

On Mon, Feb 9, 2015 at 7:39 AM, Chuck Hays <notifications@github.com> wrote:

> Anyone want to take a look at this?
>
> —
> <https://github.com/webrtc/apprtc/pull/46#issuecomment-73529380>.
>



-- 
Jiayang Liu

chuckhays:
Makes sense to me, I removed the updates to /message and kept the test fixes, the combining of constants, and lint fixes.

chuckhays:
Merged with @tkchin

jiayliu:
lgtm

jiayliu:
Please update the PR description

tkchin:
lgtm

PR #49, opened 2015-02-06
Added manifest.json for Add to Homescreen

samdutton:
Changes to enable [Add to Homescreen](https://developer.chrome.com/multidevice/android/installtohomescreen).

### DIFF
--- a/src/web_app/html/index_template.html
+++ b/src/web_app/html/index_template.html
@@ -22,6 +22,7 @@
 
   <title>AppRTC</title>
 
+  <link rel="manifest" href="manifest.json">
   <link rel="icon" sizes="192x192" href="/images/webrtc-icon-192x192.png">
 {% if not chromeapp %}
   <link rel="canonical" href="{{ room_link }}">
--- a/src/web_app/html/manifest.json
+++ b/src/web_app/html/manifest.json
@@ -1,0 +1,15 @@
+{
+  "short_name": "AppRTC",
+  "name": "WebRTC reference app",
+  "icons": [
+    {
+      "src": "/images/webrtc-icon-192x192.png",
+      "sizes": "192x192",
+      "type": "image/png",
+      "density": "4.0"
+    }
+  ],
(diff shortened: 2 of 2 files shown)


juberti:
lgtm

samdutton:
@marwahvikas we'll need to deploy this.

I'll test once it's live.

andresusanopinto:
This is cool!

samdutton:
BTW - manifest.json doesn't seem to available -
apprtc.appspot.com/manifest.json gives a 404.

Vikas - could you deploy this? The file is here
<https://github.com/webrtc/apprtc/blob/master/src/web_app/html/manifest.json>
on the repo.

If anyone has requests for 'theme-color' - the colour of the bar at the top
of the app tab - let me know :). At the moment it's white. Maybe we could
randomise it :).

marwahvikas:
Sure Sam i can do that.

marwahvikas:
Sam, chatted with Jiayang don't you need to add the handler for manifest.json before it can be deployed?

samdutton:
Ah - forgot that. I've just done a commit which adds a .json handler to
app.yaml.

marwahvikas:
Thanks Sam will deploy.

src/web_app/html/index_template.html
```diff
@@ -22,6 +22,7 @@
 
   <title>AppRTC</title>
 
+  <link rel="manifest" href="manifest.json">
```
KaptenJansson: @samdutton could this be cause of #181 ? If it's for some reason requested when the browser is on https://apprtc.appspot.com/r/roomname, would that result it the base URL being `https://apprtc.appspot.com/r/`rather than `https://apprtc.appspot.com/` ?

samdutton:
> @samdutton could this be cause of #181

Yes — that makes sense.

Changing the manifest href should cure this: `<link rel="manifest" href="/manifest.json">`.

I'll do a PR.

PR #50, opened 2015-02-06
Rewrite build_app_engine_package using python

jiayliu:
To make it possible to build on Windows.


jiayliu:
@phoglund

jiayliu:
@tkchin

Gruntfile.js
```diff
@@ -58,15 +58,8 @@ module.exports = function(grunt) {
...
-      buildVersion: {
-        command: './build/build_version_file.sh',
-        options: {
-          stdout: true,
-          stderr: true
-        }
-      },
       buildAppEnginePackage: {
-        command: './build/build_app_engine_package.sh',
+        command: 'python ./build/build_app_engine_package.py src out/app_engine',
```
tkchin: Can we put out/app_engine into a variable? It's used several times in the file.

build/build_app_engine_package.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/python
```
tkchin: nit: run gpylint
I don't know how strict we want to be with build files, but seems to be a good thing to get into a habit of running lint tools.

build/build_app_engine_package.py
```diff
@@ -0,0 +1,62 @@
...
+            for name in files:
+                # loopback.js is not compiled by Closure and needs to be copied
+                # separately.
+                if name == 'loopback.js':
+                    dest_js_path = os.path.join(dest_path, 'js')
+                    os.makedirs(dest_js_path)
+                    shutil.copy(os.path.join(dirpath, name), dest_js_path)
+                    break
+
+    os.system('build/build_version_file.sh')
```
tkchin: Does this shell file, and the other shell files, also need to be converted to Python for use in Windows?

jiayliu: Patrik is working on converting the python test script.
build_version_file is not required to run the browser tests.

PTAL

On Fri, Feb 6, 2015 at 4:15 PM, Zeke Chin <notifications@github.com> wrote:

> <https://github.com/webrtc/apprtc/pull/50#discussion_r24282451>:
>
> > +        elif dirpath.endswith('app_engine'):
> > +            for name in files:
> > +                if name.endswith('.py') and name.find('test') == -1 \
> > +                    or name.endswith('.yaml'):
> > +                    shutil.copy(os.path.join(dirpath, name), dest_path)
> > +        elif dirpath.endswith('js'):
> > +            for name in files:
> > +                # loopback.js is not compiled by Closure and needs to be copied
> > +                # separately.
> > +                if name == 'loopback.js':
> > +                    dest_js_path = os.path.join(dest_path, 'js')
> > +                    os.makedirs(dest_js_path)
> > +                    shutil.copy(os.path.join(dirpath, name), dest_js_path)
> > +                    break
> > +
> > +    os.system('build/build_version_file.sh')
>
> Does this shell file, and the other shell files, also need to be converted
> to Python for use in Windows?
>
> —
> <https://github.com/webrtc/apprtc/pull/50/files#r24282451>.
>



-- 
Jiayang Liu

tkchin:
lgtm % travis

PR #51, opened 2015-02-07
Rewrite build_version_file in python

jiayliu:


jiayliu:
@phoglund

build/build_app_engine_package.py
```diff
@@ -15,6 +17,39 @@
...
+
+def build_version_info_file(dest_path):
+  """Build the version info JSON file."""
+  version_info = {
+      'gitHash': None,
+      'time': None,
+      'branch': None
+  }
+
+  lines = call_cmd_and_return_output_lines(['git', 'log', '-1'])
```
tkchin: try except? Could get OSError if git isn't installed.

build/build_app_engine_package.py
```diff
@@ -15,6 +17,39 @@
...
+def call_cmd_and_return_output_lines(cmd):
+  process = subprocess.Popen(cmd, stdout=subprocess.PIPE)
+  output = process.communicate()[0]
+  return output.split('\n')
+
+
+def build_version_info_file(dest_path):
+  """Build the version info JSON file."""
+  version_info = {
+      'gitHash': None,
```
tkchin: Think it's better to not initialize and let block below add entries. Don't know what happens when json.dumps gets a None value.

build/build_app_engine_package.py
```diff
@@ -15,6 +17,39 @@
...
+    if version_info['gitHash'] is not None and version_info['time'] is not None:
+      break
+
+  lines = call_cmd_and_return_output_lines(['git', 'branch'])
+  for line in lines:
+    if line.startswith('*'):
+      version_info['branch'] = line.partition(' ')[2].strip()
+      break
+
+  with open(dest_path, 'w') as f:
```
tkchin: Safer to either check dest_path exists (even though it does because it's being called from main) or handle potential not found exception here.

build/build_app_engine_package.py
jiayliu: It will be KEY:null

build/build_app_engine_package.py
jiayliu: Done. PTAL

On Fri, Feb 6, 2015 at 5:34 PM, Zeke Chin <notifications@github.com> wrote:

> <https://github.com/webrtc/apprtc/pull/51#discussion_r24284985>:
>
> > +  lines = call_cmd_and_return_output_lines(['git', 'log', '-1'])
> > +  for line in lines:
> > +    if line.startswith('commit'):
> > +      version_info['gitHash'] = line.partition(' ')[2].strip()
> > +    elif line.startswith('Date'):
> > +      version_info['time'] = line.partition(':')[2].strip()
> > +    if version_info['gitHash'] is not None and version_info['time'] is not None:
> > +      break
> > +
> > +  lines = call_cmd_and_return_output_lines(['git', 'branch'])
> > +  for line in lines:
> > +    if line.startswith('*'):
> > +      version_info['branch'] = line.partition(' ')[2].strip()
> > +      break
> > +
> > +  with open(dest_path, 'w') as f:
>
> Safer to either check dest_path exists (even though it does because it's
> being called from main) or handle potential not found exception here.
>
> —
> <https://github.com/webrtc/apprtc/pull/51/files#r24284985>.
>



-- 
Jiayang Liu

tkchin:
lgtm

PR #52, opened 2015-02-09
Rewriting python test provisioning from .sh to .py, separating travis and dev workflows

phoglund:
This patch gets rid of the .sh file we used to download dependencies for the python tests. This should now work on Windows.

Furthermore, I try to improve the confusing setup around webtests. The script will no longer attempt to autoinstall webtest since it requires easy_install - it now tells the developer what to do and exits cleanly. I could have made the script download and install easy_install, but I don't like running sudo on the developer's behalf. Adding more of that seems wrong.

Travis will still need to run sudo and it already has easy_install/setuptools, so I create a new workflow for Travis where this is done automatically.

### DIFF
--- a/build/build_app_engine_package.py
+++ b/build/build_app_engine_package.py
@@ -57,7 +57,7 @@ def build_version_info_file(dest_path):
     print str(e)
 
 
-def main(src_path, dest_path):
+def CopyApprtcSource(src_path, dest_path):
   if os.path.exists(dest_path):
     shutil.rmtree(dest_path)
   os.makedirs(dest_path)
@@ -80,7 +80,8 @@ def main(src_path, dest_path):
           shutil.copy(os.path.join(dirpath, name), dest_html_path)
     elif dirpath.endswith('app_engine'):
       for name in files:
-        if (name.endswith('.py') or name.endswith('.yaml')):
+        if (name.endswith('.py') and 'test' not in name
+            or name.endswith('.yaml')):
           shutil.copy(os.path.join(dirpath, name), dest_path)
     elif dirpath.endswith('js'):
       for name in files:
@@ -94,11 +95,28 @@ def main(src_path, dest_path):
 
   build_version_info_file(os.path.join(dest_path, 'version_info.json'))
 
-if __name__ == '__main__':
+
+def CopyTests(src_path, dest_path):
+  for dirpath, _, files in os.walk(src_path):
+    if dirpath.endswith('app_engine'):
+      tests = [name for name in files if 'test' in name]
+      for test in tests:
+        shutil.copy(os.path.join(dirpath, test), dest_path)
+
+
+def main():
   parser = optparse.OptionParser(USAGE)
-  _, args = parser.parse_args()
+  parser.add_option("-t", "--include-tests", action="store_true",
+                    help='Also copy python tests to the out dir.')
+  options, args = parser.parse_args()
   if len(args) != 2:
     parser.error('Error: Exactly 2 arguments required.')
 
   src_path, dest_path = args[0:2]
-  main(src_path, dest_path)
+  CopyApprtcSource(src_path, dest_path)
+  if options.include_tests:
+    CopyTests(src_path, dest_path)
+
+
+if __name__ == '__main__':
+  sys.exit(main())
--- a/build/build_app_engine_package.py
+++ b/build/build_app_engine_package.py
@@ -10,6 +10,8 @@ import shutil
 import subprocess
 import sys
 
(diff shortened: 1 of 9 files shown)


build/build_app_engine_package.py
```diff
@@ -80,8 +80,7 @@ def main(src_path, dest_path):
           shutil.copy(os.path.join(dirpath, name), dest_html_path)
     elif dirpath.endswith('app_engine'):
       for name in files:
-        if (name.endswith('.py') and name.find('test') == -1 or
```
phoglund: It seemed hackish to me to just copy the tests when they're about to run. This will always copy them. Any reason why this would be a problem?

jiayliu: Everything in out/app_engine is pushed to the server during deployment. In general we don't want to push unnecessary files.

phoglund:
PTAL! I now copy the tests in the build apprtc package script, that seemed to be where that logic belonged. I also made copying tests a separate function since the existing function is already very complicated.

phoglund:
ping

build/build_app_engine_package.py
```diff
@@ -95,14 +95,28 @@ def main(src_path, dest_path):
...
-    sys.exit(1)
+    parser.error('Error: Exactly 2 arguments required.')
 
-  SRC_PATH = args[0]
-  DEST_PATH = args[1]
-  main(SRC_PATH, DEST_PATH)
+  src_path, dest_path = args[0:2]
+  CopyApprtcSource(src_path, dest_path)
+  if options.include_tests:
+    CopyTests(src_path, dest_path)
```
jiayliu: The tests are copied but never cleaned up, so if someone run the python tests once, the test files will remain in the out folder, and it's error prone to push all of them to GAE.

phoglund: So you mean we should remove the tests from the out folder after running them? That sounds pretty ridiculous, and it would be error-prone as well; consider for instance the case when the test crashes half-way through. We would have to be very careful to ensure cleanup in that case.

All the pushing user needs to do is to run grunt build, which will wipe and re-assemble the out folder. That will remove the tests. In fact, we could even make pushing a grunt action so the user don't even have to remember this. Like, grunt push = grunt build + a grunt action which calls dev_appserver.py upload ... We do already have a GAE sdk checked out and Grunt knows where it is, so that should be doable.

Before we do all that, let me challenge your assertion that pushing the tests is super bad. Why? Because it takes a few more kb of uploading? The tests won't get run on the server anyway.

jiayliu: Pushing the tests is like pushing any arbitrary file not needed, and we
should avoid it with reasonable effort. Removing the test files in the
successful case seems reasonable effort to me.

I don't think we need to worry about cleanup when the test crashes, since
one should not push until all tests pass.

I'm not sure about the "grunt push" idea, for
1. we also need to support "grunt deploy" for local deployment to be
consistent.
2. we need to support all the push/deploy options.


On Mon, Feb 16, 2015 at 12:12 AM, phoglund <notifications@github.com> wrote:

> <https://github.com/webrtc/apprtc/pull/52#discussion_r24735967>:
>
> >
> > -  SRC_PATH = args[0]
> > -  DEST_PATH = args[1]
> > -  main(SRC_PATH, DEST_PATH)
> > +  src_path, dest_path = args[0:2]
> > +  CopyApprtcSource(src_path, dest_path)
> > +  if options.include_tests:
> > +    CopyTests(src_path, dest_path)
>
> So you mean we should remove the tests from the out folder after running
> them? That sounds pretty ridiculous, and it would be error-prone as well;
> consider for instance the case when the test crashes half-way through. We
> would have to be very careful to ensure cleanup in that case.
>
> All the pushing user needs to do is to run grunt build, which will wipe
> and re-assemble the out folder. That will remove the tests. In fact, we
> could even make pushing a grunt action so the user don't even have to
> remember this. Like, grunt push = grunt build + a grunt action which calls
> dev_appserver.py upload ... We do already have a GAE sdk checked out and
> Grunt knows where it is, so that should be doable.
>
> Before we do all that, let me challenge your assertion that pushing the
> tests is super bad. Why? Because it takes a few more kb of uploading? The
> tests won't get run on the server anyway.
>
> —
> <https://github.com/webrtc/apprtc/pull/52/files#r24735967>.
>



-- 
Jiayang Liu

phoglund: All right, let's go with that then. I added some scripts that clean up the tests.

phoglund:
PTAL

jiayliu:
Thx!
lgtm

PR #53, opened 2015-02-10
Generate and return a session id to client.

chuckhays:
Generates a uuid for use as the client_id on the client side. Used for:
* /message
* /leave (only clients for open rooms, direct rooms still use gcm id)
* wss messaging

The gcm id still serves as the primary id on the server side, but client objects can be looked up via the temporary session id.

@tkchin @jiayliu


src/app_engine/room.py
```diff
@@ -54,6 +54,11 @@ def get_occupancy(self):
   def has_client(self, client_id):
     return client_id in self.clients
 
+  def has_client_by_session_id(self, client_session_id):
```
jiayliu: Add tests for the new methods?

src/app_engine/client.py
```diff
@@ -2,25 +2,35 @@
...
-"""Client model
+"""Client model.
 
 This module implements the client data model.
 """
 
+import uuid
+
+
 # For now we have (room_id, client_id) pairs are 'unique' but client_ids are
```
jiayliu: Please update the comment and explain the difference between client_id and session_id.

tkchin:
Side note, but I think we're starting to pay the price of supporting two room models in essentially the same model. It seems overkill that we have a notion of client_id, session_id, room_id and gcm_id, and that we overload the same handler to take multiple types of ids. I'm okay with adding session_id for now but I think we should look into refactoring so that the different code paths (open vs direct) are structured more cleanly.

chuckhays:
@tkchin I agree, this change took much longer than it needed to because of that complexity. I'm still not entirely happy with how it ended up. But we are learning useful info about how we could restructure it at least.

chuckhays:
Added tests and updated comment.

src/app_engine/client.py
```diff
@@ -2,25 +2,38 @@
...
+
+
+# For OPEN type rooms, client_id is a generated random number.
+# For DIRECT type rooms, client_id is the device's gcm id.
+# (room_id, client_id) pairs are 'unique', but in the OPEN case, client_id
+# values may not be. This uniqueness is not enforced however and bad things
+# may happen if RNG generates non-unique numbers for client_ids. We also
+# have a special loopback client id.
+# The session_id value is an uuid. It is used by clients as the identifier
+# to the WSS server as well as for calling /leave in an OPEN type room.
```
jiayliu: And for /message right?

chuckhays: You are correct, I wrote WSS for the messaging portion for forgot the /message calls, good catch.

PR #56, opened 2015-02-10
fix main.css link in params.html

jiayliu:


chuckhays:
LGTM.

src/web_app/html/params.html
```diff
@@ -89,8 +89,5 @@
     <a href="//github.com/GoogleChrome/webrtc" title="View the repository" id="viewSource">github.com/GoogleChrome/webrtc</a>
 
   </div>
-
```
andresusanopinto: @jiayliu 
Was this intentional?
Why is it missing from the pull request description?

jiayliu: Oops, I didn't realize it pointed to the ga.js in samples. Will add it back.

jiayliu: But wait, why do we need it? This is page is not in samples. @andresusanopinto

andresusanopinto: I have no clue if it is needed or not, and did not spend any brain on it.. i just happen to be browsing changes (when apprtc was not working) and though it was strange a fix for .css path touched this line.

PR #58, opened 2015-02-11
Fixed a typo in message_page.py

wzh--:
A typo that causes the script to stop.


wzh--:
@chuckhays @jiayliu PTAL

chuckhays:
LGTM.

PR #62, opened 2015-02-11
Delay usage of AppAssertionCredentials

milkypostman:
This should avoid some import problems on the test servers related to
issue #21.

### DIFF
--- a/src/app_engine/analytics.py
+++ b/src/app_engine/analytics.py
@@ -13,7 +13,7 @@ import httplib2
 import webapp2
 from apiclient import discovery
 from google.appengine.api import app_identity
-from oauth2client.appengine import AppAssertionCredentials
+import oauth2client.appengine
 import oauth2client.client
 
 import constants
@@ -24,7 +24,7 @@ class Analytics(object):
 
   All data is streamed to BigQuery.
 
-"""
+  """
   def __init__(self):
     is_running_locally = os.environ.get('APPLICATION_ID', '').startswith('dev')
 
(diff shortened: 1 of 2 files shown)


phoglund:
Where does the secrets.json normally come from? If I clone this repo and run grunt build, will the secrets.json be there or does it need to be copied in as a separate step?

milkypostman:
@phoglund you have to run `grunt build` and then all that stuff is built.

previously you could run it directly from the appengine directory (sans javascript being compiled with `grunt build`) but now you have to build all of it.

chuckhays:
LGTM.

phoglund:
Ok. My point is we need some way to bring up the server on the bots without importing the stuff that broke the bots before (e.g SignedJwtSomething). The bots do build using grunt build. Will this patch solve that or do we need to add a app.yaml flag the bots can override?

milkypostman:
@phoglund how are you running the server on the bots? If you are using `dev_appserver.py` then it should just work. The problem was the import and now the actual object that was being imported should never be used. When running `dev_appserver.py` we look for `secrets.json` and if it's not there then we just do logging to stdout. The only time this would fail is if it was running *on* AppEngine.

phoglund:
It does use dev_appserver, but the problem was the crypto libraries were too old on the bot, but not on our workstations.

To test this, go into the google_appengine SDK you're launching from. Edit google_appengine/oauth2client/client.py. Go to line 44 or thereabouts and change the line HAS_CRYPTO = True to HAS_CRYPTO = False. Run your code with google_appengine/dev_appserver.py path/to/checkout/out/app_engine and ensure you can still get a call up on that server.

If that still works you're good. Remember to change back the HAS_CRYPTO line later :)

milkypostman:
It works. Since it's very late here I will merge this tomorrow when I get in.

PR #63, opened 2015-02-11
Add the AnalyticsPage to AppRTC.

milkypostman:
Implements functionality to handle an Analytics request from the client.

@jiayliu @tkchin @chuckhays  I implemented this with immediate needs in mind but I considered the problem of caching on the client and wanting to send up multiple events in a single request. However, after I built that version I realized that we have to be considerate of the response that the server gives back. When we only send one event we know that a failure means that single event insertion failed. However, if we allow multiple events per request we have complexity in the response as to which events were OK and which weren't. So this instance can be extended in the future if we feel we need that added complexity.


src/app_engine/apprtc.py
```diff
@@ -546,6 +546,110 @@ def get(self):
     params = get_room_parameters(self.request, None, None, None)
     self.response.write(json.dumps(params))
 
+class AnalyticsPage(webapp2.RequestHandler):
```
jiayliu: Move this to its own file?

src/app_engine/apprtc_test.py
```diff
@@ -125,5 +143,122 @@ def testCallerMessagesForwardedToCallee(self):
     self.makePostRequest('/leave/' + room_id + '/' + caller_id)
     self.makePostRequest('/leave/' + room_id + '/' + callee_id)
 
+  def testAnalyticsPage(self):
```
jiayliu: Move to its own file, to minimize conflict when merging to the other branch?

src/app_engine/apprtc.py
```diff
@@ -546,6 +546,110 @@ def get(self):
...
+  """
+  def _write_response(self, result, message=None):
+    response = {'result': result}
+    if message is not None:
+      response['message'] = message
+
+    self.response.write(json.dumps(response));
+
+  def _write_invalid_request(self):
+    self.error(400)
```
jiayliu: The other GAE APIs does not return an error code except for server internal error. Can you just return 200?

src/app_engine/apprtc.py
```diff
@@ -546,6 +546,110 @@ def get(self):
...
+    return time.time()
+
+  def post(self):
+    try:
+      msg = json.loads(self.request.body)
+    except ValueError:
+      return self._write_invalid_request()
+
+    valid_request = False
+    type = msg.get('type')
```
jiayliu: Move these string literals to constants?

src/app_engine/apprtc.py
milkypostman: That's not completely true as there is a place we return a 500. I'd argue this is the *right* way to do error handling and so we should move towards.

chuckhays: +1 I would prefer to use http status codes where appropriate.

src/app_engine/apprtc.py
milkypostman: done

src/app_engine/apprtc_test.py
milkypostman: done

src/app_engine/apprtc.py
milkypostman: done

milkypostman:
@jiayliu @tkchin @chuckhays would you mind providing additional comments or lgtm? I think I have addressed previous comments.

jiayliu:
lgtm

src/app_engine/analytics_page.py
```diff
@@ -0,0 +1,117 @@
...
+
+  Currently the AnalyticsPage handles the following types:
+  'event': A JSON event object of the form:
+          {
+            'event_type': <string corresponding to an attribute of EventType>,
+            'event_time_ms': <client time when the event occurred>,
+            'room_id': <id of the room corresponding to the event [optional]>,
+          }
+
+      The event is reported to Analytics.report_event with the
```
tkchin: Can this be clarified? I read this several times and still didn't quite understand what offset you're calculating and for what purposes the timestamps are stored. When you say client time do you mean client local time? Or client UTC? Is the offset used to increase accuracy of reported time in some way to adjust for clock sync? Why do you need the original event time?

src/app_engine/analytics_page.py
```diff
@@ -0,0 +1,117 @@
...
+      adjusted by the calculated offset. The original 'event_time_ms'
+      is reported as the 'client_timestamp'. This ignores the latency
+      of opening a connection and sending the body of the message to
+      the Analytics page. Travel time should be negligible.
+  """
+
+  def _write_response(self, result, message=None):
+    response = {'result': result}
+    if message is not None:
+      response['message'] = message
```
tkchin: Should this be a dictionary stuffed into a field, or should the keys be present in response directly? What would make sense from a proto perspective?

src/app_engine/analytics_page.py
```diff
@@ -0,0 +1,117 @@
...
+
+    self.response.write(json.dumps(response))
+
+  def _write_invalid_request(self):
+    self.error(400)
```
tkchin: In other handlers we've been writing bad requests into the response['result'] field and only writing HTTP codes for 500. The reason for that is that we've been providing facilities for why exactly the request was rejected - eg invalid callee id as opposed to invalid room id. Would be good to keep this consistent across APIs.

src/app_engine/apprtc.py
```diff
@@ -546,6 +547,7 @@ def get(self):
     params = get_room_parameters(self.request, None, None, None)
     self.response.write(json.dumps(params))
 
+
 app = webapp2.WSGIApplication([
     ('/', MainPage),
     ('/leave/(\w+)/(\w+)', LeavePage),
```
tkchin: nit: alphabetize handlers

src/app_engine/constants.py
```diff
@@ -42,6 +42,21 @@ class EventType(object):
   # Event signifying that a room enters the state of having exactly
   # two participants.
   ROOM_SIZE_2 = 'room_size_2'
+  ICE_CONNECTION_STATE_CONNECTED = 'ice_connection_state_connected'
+
+
+class AnalyticsField(object):
```
tkchin: Are you planning to add methods here, or do these exist just for namespacing? If the latter, can these be constants instead?
Also, can this be placed in analytics.py instead?

src/app_engine/analytics_page.py
```diff
@@ -0,0 +1,117 @@
...
+    if not valid_request:
+      self._write_invalid_request()
+    else:
+      self._write_response(constants.RESPONSE_SUCCESS)
+
+    return
+
+  def _handle_event(self, content, request_time_ms):
+    try:
+      imp = json.loads(content)
```
tkchin: what is imp?

src/app_engine/constants.py
milkypostman: Are you suggesting something like `ANALYTICS_FIELD_TYPE` or something?

I am happy to move it to analytics. In that case we way want to move LogField too; I know you suggested this before but we never resolved the issue.

src/app_engine/analytics_page.py
milkypostman: But why not use 400 to indicate "Bad Request" as the HTTP spec says: http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html?

src/app_engine/constants.py
tkchin: Yup ANALYTICS_FIELD_TYPE or something. Should be same number of characters. Keeping consistent with the rest of our constants.

src/app_engine/analytics_page.py
tkchin: If this was the only API in the absence of our call APIs it would be fine. I just want to avoid this situation where clients have to handle 400 for analytics but INVALID_CALLEE_ID for call APIs.

src/app_engine/analytics_page.py
milkypostman: yeah, should be event.

src/app_engine/analytics_page.py
milkypostman: so is updating the other APIs even an option or is the previous decision influence everything going forward?

tkchin: If we update the other APIs we'll still want invalid request reason for them. So the response code would be 400 but we'd still have JSON response content with the specific error code. This seems to just add client / server parsing code unnecessarily.

milkypostman: Oh, sorry, I don't care about the content sent back; happy to change that to whatever is easiest. I was addressing your comment about sending the 500 code.

milkypostman: now that I read this thread again, maybe i don't understand what you're proposing.

src/app_engine/analytics_page.py
milkypostman: i don't think we should stuff dictionaries into other dictionaries in general. from a proto perspective you would have a Request proto and Response proto and you would know to decode that specifically. And while you can do this wrapping i don't think it's great practice.

but this makes me think I am doing some redundant wrapping in my request to the analytics page which I should avoid. There is no reason to have a "content" field that you then need to unencode.

src/app_engine/analytics_page.py
tkchin: proposal is: Instead of 400 code, send 200 with JSON response { 'result': 'invalid' } since that's already being sent back and checked.

src/app_engine/constants.py
milkypostman: I am leaving these as-is because I want the ability to generate enums for JS and I want them to be namespaced there. It seems easier to me, and cleaner, to represent these as classes in Python to make the generation easier and provide some similarity to the JS counterparts. I did move them to the analytics and analytics_page modules as appropriate.

src/app_engine/apprtc.py
milkypostman: Done

src/app_engine/analytics_page.py
milkypostman: Can you take a look at the new wording. I reworked how the JSON request is sent as well to make it more "proto" like based on your other comments below.

src/app_engine/analytics_page.py
milkypostman: Changed this in the most recent version.

src/app_engine/analytics_page.py
milkypostman: Done.

milkypostman:
@tkchin PTAL

tkchin:
Looking now, sorry for delay.

tkchin:
lgtm

PR #65, opened 2015-02-12
Add a websocket read timeout to temporarily fix socket leak

jiayliu:


jiayliu:
Zeke, PTAL

tkchin:
lgtm

This looks right given we had errors on 443. Is it possible to write a stress test unit test to check the fix? Probably good to get this in so we can fix error, but if it doesn't take long to write might as well add it.

jiayliu:
The real problem is found and fixed in the new patch. PTAL

milkypostman:
LGTM.

tkchin:
sweet, lgtm

PR #67, opened 2015-02-12
Fix response for message handler when GAE forwards messages to collider.

tkchin:
When GAE forwards messages to collider currently it writes response twice.


tkchin:
@jiayliu @chuckhays PTAL

chuckhays:
LGTM.

tkchin:
TBR @jiayliu

PR #73, opened 2015-02-13
Factor out Google API auth.

milkypostman:
This takes the portion of the analytics code we use to create an
authorized API object and puts it in a shared googleapi module. We can
then use this later for any cloud API.

### DIFF
--- a/src/app_engine/analytics.py
+++ b/src/app_engine/analytics.py
@@ -1,7 +1,8 @@
 # Copyright 2015 Google Inc. All Rights Reserved.
 
+"""Module for pushing analytics data to BigQuery."""
+
 import datetime
-import json
 import logging
 import os
 import sys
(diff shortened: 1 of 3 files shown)


src/app_engine/analytics.py
```diff
@@ -26,39 +21,22 @@ class Analytics(object):
...
-
-  def _build_bigquery_object(self, credentials):
-    http = credentials.authorize(httplib2.Http())
-    return discovery.build("bigquery", "v2", http=http)
+    self.bigquery = self._build_bigquery_object()
+    if self.bigquery is None:
+      logging.warning('Unable to build BigQuery API object. Logging disabled.')
+
+  def _build_bigquery_object(self):
+    return googleapi.build(scope=constants.BIGQUERY_URL,
```
jiayliu: nit: googleapi may be confusing since it sounds like a google provided module,  like google.appengnie.api. How about a more specific name, like authutils?

tkchin:
Could you run gpylint on the files?

milkypostman:
@jiayliu I changed it to `apiauth` as I thought `utils` is a bit vague anyways. i actually think this function belongs in a more general helper or apihelper.py maybe? I dunno. naming is the pits.

@tkchin ran the linter. Thanks. I didn't even think about that.

jiayliu:
lgtm

PR #74, opened 2015-02-13
Fix bad response for message handler when GAE forwards messages to collider.

tkchin:
Porting fix from:
40d185b942801dc712158e1657eeb66479cdbeb4


tkchin:
@jiayliu @chuckhays PTAL

jiayliu:
lgtm

PR #76, opened 2015-02-13
Compute engine handle page.

milkypostman:
Adds a page that can be used to start the compute instances programmatically.

@jiayliu  :: *Need to add unit tests but wanted to get your take on this idea when you get time.*

### DIFF
--- a/src/app_engine/compute_page.py
+++ b/src/app_engine/compute_page.py
@@ -1,0 +1,104 @@
+# Copyright 2015 Google Inc. All Rights Reserved.
+
+"""Compute page for handling tasks related to compute engine."""
+
+import logging
+
+import webapp2
+
+import apiauth
+from google.appengine.api import app_identity
+from google.appengine.api import taskqueue
+
+
+# Page actions
+# Get the status of an instance.
+ACTION_STATUS = 'status'
+
+# Start the compute instance if it is not already started. When an
+# instance is in the TERMINATED state, it will be started. If the
+# instance is already RUNNING do nothing. If the instance is in an
+# intermediate state--PROVISIONING, STAGING, STOPPING--the task is
+# requeued.
+ACTION_START = 'start'
+
+# Constants for the Compute Engine API
+COMPUTE_STATUS = 'status'
+COMPUTE_STATUS_TERMINATED = 'TERMINATED'
+COMPUTE_STATUS_RUNNING = 'RUNNING'
+
+# Seconds between API retries.
+TASK_RETRY_S = 3
+
+COMPUTE_API_URL = 'https://www.googleapis.com/auth/compute'
+
+
+class ComputePage(webapp2.RequestHandler):
+  """Page to handle requests against GCE."""
+
+  def __init__(self, request, response):
+    # Call initialize rather than the parent constructor fun. See:
+    # https://webapp-improved.appspot.com/guide/handlers.html#overriding-init
+    self.initialize(request, response)
+
+    self.compute_service = self._build_compute_service()
+    if self.compute_service is None:
+      logging.warning('Unable to create Compute service object.')
+
+  def _build_compute_service(self):
+    return apiauth.build(scope=COMPUTE_API_URL,
+                         service_name='compute',
+                         version='v1')
+
+  def _enqueue_start_task(self, instance, zone):
+    taskqueue.add(url='/compute/start/%s/%s' % (instance, zone),
+                  countdown=TASK_RETRY_S)
+
+  def _maybe_start_instance(self, instance, zone):
+    """Implementation for start action.
+
+    Args:
+      instance: Name of the instance to start.
+      zone: Name of the zone the instance belongs to.
+
+    """
+    if self.compute_service is None:
+      logging.warning('Unable to start Compute instance, service unavailable.')
+      return
+
+    status = self._compute_status(instance, zone)
+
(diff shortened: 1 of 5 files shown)


src/app_engine/compute_page.py
```diff
@@ -0,0 +1,86 @@
...
+    if ComputePage.compute_service is None:
+      logging.warning('Unable to create Compute service object.')
+
+  def _build_compute_service(self):
+    return apiauth.build(scope=constants.COMPUTE_URL,
+                         serviceName='compute',
+                         version='v1')
+
+  def _enqueue_start(self):
+    taskqueue.add(url='/compute/start',
```
jiayliu: Can we add some authentication info so that an external user cannot do it, but only GAE can issue it? It makes me feel uneasy to allow anyone to manipulate the VM instance.

milkypostman: I added a 'admin' login required for these urls now. This should secure things so that the tasks can run them.  I hadn't seen this: https://cloud.google.com/appengine/docs/python/taskqueue/overview-push#Python_Securing_URLs_for_tasks

jiayliu: The idea looks good to me. Thanks!

On Fri, Feb 13, 2015 at 5:00 PM, Donald Ephraim Curtis <

> In src/app_engine/compute_page.py
> <https://github.com/webrtc/apprtc/pull/76#discussion_r24709530>:
>
> > +  compute_service = None
> > +
> > +  def __init__(self, request, response):
> > +    self.initialize(request, response)
> > +    if ComputePage.compute_service is None:
> > +      ComputePage.compute_service = self._build_compute_service()
> > +    if ComputePage.compute_service is None:
> > +      logging.warning('Unable to create Compute service object.')
> > +
> > +  def _build_compute_service(self):
> > +    return apiauth.build(scope=constants.COMPUTE_URL,
> > +                         serviceName='compute',
> > +                         version='v1')
> > +
> > +  def _enqueue_start(self):
> > +    taskqueue.add(url='/compute/start',
>
> I added a 'admin' login required for these urls now. This should secure
> things so that the tasks can run them. I hadn't seen this:
> https://cloud.google.com/appengine/docs/python/taskqueue/overview-push#Python_Securing_URLs_for_tasks
>
> —
> <https://github.com/webrtc/apprtc/pull/76/files#r24709530>.
>



-- 
Jiayang Liu

jiayliu:
Looks really good. Please add tests.

milkypostman:
@tkchin ready for comments. some cleanup needs to be done with common code but I am going to save that for a later commit. mainly common code shared by tests. #93 tracks this.

milkypostman:
WAIT! Let me lint this thing!

milkypostman:
OK linted.

src/app_engine/compute_page.py
```diff
@@ -0,0 +1,98 @@
...
+COMPUTE_API_URL = 'https://www.googleapis.com/auth/compute'
+
+
+class ComputePage(webapp2.RequestHandler):
+  """Page to handle requests against GCE."""
+
+  compute_service = None
+
+  def __init__(self, request, response):
+    self.initialize(request, response)
```
tkchin: Clever. I had no idea you could do this, but makes sense.
Do you do self.initialize or would it be better to do super(webapp2.RequestHandler, self).__init__?

tkchin: autoformat woes
_ _ i n i t _ _

milkypostman: yeah i would normally do that but I read this: https://webapp-improved.appspot.com/guide/handlers.html#overriding-init

apparently that's the way they prescribe but yeah, I started out with calling the superclass function. maybe they think this is more explicit looking and less magic?

src/app_engine/compute_page.py
```diff
@@ -0,0 +1,98 @@
...
+
+    logging.info('GCE VM \'%s (%s)\' status: \'%s\'.',
+                 instance, zone, status)
+
+    if status == COMPUTE_STATUS_RUNNING:
+      return
+
+    if status == COMPUTE_STATUS_TERMINATED:
+      logging.info('Starting GCE VM: %s (%s)', instance, zone)
+      ComputePage.compute_service.instances().start(
```
tkchin: Not familiar with how discovery services work - I'm guessing that's what's happening here? What's being executed?
Maybe add comments about how compute instance is being started at top of file.

src/app_engine/compute_page.py
tkchin: Weird. Add comment?

src/app_engine/compute_page.py
```diff
@@ -0,0 +1,98 @@
...
+          project=app_identity.get_application_id(),
+          instance=instance,
+          zone=zone).execute()
```
jiayliu: Why not just return from here?

milkypostman: because we want to make sure that the task starts so we want to queue the task again. i'm actually not sure how this is going to work in practice. I may need to actually name the task in the case where the prober fails again and fires off another task.....

hrm, need to think about this more.

milkypostman:
Ready for another look. Comments should be addressed now. I separated the tests for different instance states to separate test functions. This is a bit cleaner.

jiayliu:
lgtm

tkchin:
lgtm!

PR #77, opened 2015-02-14
Remove error log from Collider status report

jiayliu:
The error message may contain IP address and should not be made public.


jiayliu:
TBR= @tkchin

PR #79, opened 2015-02-18
Move EventType LogField to the analytics module.

milkypostman:
@tkchin I thought about moving these to constants but the issue is that later I am going to need to generate JS based on some of these enums and I prefer they be namespaced somehow. I think that just making a class to encapsulate them is as good of approach as any.

### DIFF
--- a/src/app_engine/analytics_test.py
+++ b/src/app_engine/analytics_test.py
@@ -8,7 +8,6 @@ import apprtc
 import datetime
 import json
 import time
-from constants import LogField
 from google.appengine.api import memcache
 from google.appengine.ext import testbed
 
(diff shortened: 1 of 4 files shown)


jiayliu:
lgtm

src/app_engine/analytics_test.py
```diff
@@ -91,8 +90,8 @@ def tearDown(self):
   def testOnlyEvent(self):
     event_type = 'an_event'
     logDict = self.create_log_dict(
-        {LogField.TIMESTAMP: '{0}'.format(self.now_isoformat()),
-         LogField.EVENT_TYPE: event_type})
+        {analytics.LogField.TIMESTAMP: '{0}'.format(self.now_isoformat()),
```
tkchin: you can probably do from analytics import LogField here

tkchin:
lgtm

PR #80, opened 2015-02-18
Extract test utility functions to separate module and minor linting issues.

milkypostman:
I fixed some linting issues in this PR based on @tkchin suggestions in another PR. I also extracted the test_util module from previous unit tests here. This should make the AnalyticsPage PR a bit smaller and less distracting.


milkypostman:
I did separate commits in case we need to rollback one or the other, in which case we don't need to roll them *both* back.

jiayliu:
lgtm

tkchin:
lgtm

Issue #59, opened 2015-02-11
Promise polyfill in testpolyfills can't chain


chuckhays:
Our implementation of Promise in testpolyfills doesn't chain promises. I ran into this error a while back and had to restructure the app code so unit tests would pass. I didn't realize what the cause was until now.

Example code that won't work:

var p = new MyPromise(function(resolve, reject) {
    console.log('starting timer');
    setTimeout(function() { resolve('a'); }, 2000);
  });

  p.then(function(v) {
    console.log(v);
    return 'b';
  }).then(function(s) {
    console.log(s);
  })

Only the last .then will be used because .then always returns the same promise, it should return a new promise or keep a list of deferred promises.

It might be simpler to see if we can use one of the existing promise polyfills.

KaptenJansson:
Closing due to age.

PR #82, opened 2015-02-18
Set up Chrome App shutdown flow.

chuckhays:
Makes the following changes:
* Support both sync/async hangup flow. Sync is used when web browser closes so all cleanup calls can finish before browser closes, async is used when hang up button is pressed. Chrome App can't do sync xhr, so hang up flow has to be async.
* Moves websocket ownership to background.js in the Chrome App
* Adds a queue of cleanup actions in background.js that are executed when the Chrome App window closes

Only background.js knows when the app window closes, and once it has already closed, the window can no longer do xhr or web socket requests. Therefore, background.js has to take care of the clean up calls required to end the in progress call. 

@tkchin @jiayliu 

I'm still working to see if I can unit test more of the code, but wanted to send this out to get the review started, as it is a pretty complex bit of code.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -145,14 +145,17 @@ module.exports = function(grunt) {
             'src/web_app/js/adapter.js',
             'src/web_app/js/appcontroller.js',
             'src/web_app/js/call.js',
+            'src/web_app/js/constants.js',
             'src/web_app/js/infobox.js',
             'src/web_app/js/peerconnectionclient.js',
             'src/web_app/js/roomselection.js',
+            'src/web_app/js/remotewebsocket.js',
             'src/web_app/js/sdputils.js',
             'src/web_app/js/signalingchannel.js',
             'src/web_app/js/stats.js',
             'src/web_app/js/storage.js',
             'src/web_app/js/util.js',
+            'src/web_app/js/windowport.js',
           ]
         },
         options: {
@@ -184,7 +187,7 @@ module.exports = function(grunt) {
                                         'shell:getPythonTestDeps',
                                         'shell:runPythonTests',
                                         'shell:removePythonTestsFromOutAppEngineDir']);
-  grunt.registerTask('jstests', ['closurecompiler:debug', 'jstdPhantom']);
+  grunt.registerTask('jstests', ['closurecompiler:debug', 'grunt-chrome-build', 'jstdPhantom']);
   // buildAppEnginePackage must be done before closurecompiler since buildAppEnginePackage resets out/app_engine.
   grunt.registerTask('build', ['shell:buildAppEnginePackage', 'closurecompiler:debug', 'grunt-chrome-build']);
 };
--- a/src/web_app/chrome_app/manifest.json
+++ b/src/web_app/chrome_app/manifest.json
@@ -5,7 +5,7 @@
   "manifest_version": 2,
   "app": {
     "background": {
-      "scripts": ["js/background.js"]
+      "scripts": ["js/apprtc.debug.js", "js/background.js"]
     }
   },
   "icons": {
--- a/src/web_app/js/appcontroller.js
+++ b/src/web_app/js/appcontroller.js
@@ -193,7 +193,6 @@ AppController.prototype.showRoomSelection_ = function() {
 AppController.prototype.finishCallSetup_ = function(roomId) {
   this.call_.start(roomId);
 
-  window.onbeforeunload = this.call_.hangup.bind(this.call_);
   document.onkeypress = this.onKeyPress_.bind(this);
   window.onmousemove = this.showIcons_.bind(this);
 
@@ -205,6 +204,13 @@ AppController.prototype.finishCallSetup_ = function(roomId) {
   setUpFullScreen();
 
   if (!isChromeApp()) {
+    // Call hangup with async = false. Required to complete multiple
+    // clean up steps before page is closed.
+    // Chrome apps can't use onbeforeunload.
+    window.onbeforeunload = function() {
+      this.call_.hangup(false);
+    }.bind(this);
+
     window.onpopstate = function(event) {
       if (!event.state) {
         // TODO (chuckhays) : Resetting back to room selection page not
@@ -227,7 +233,8 @@ AppController.prototype.hangup_ = function() {
   this.displayStatus_('Hanging up');
   this.transitionToDone_();
 
-  this.call_.hangup();
+  // Call hangup with async = true.
+  this.call_.hangup(true);
 };
 
 AppController.prototype.onRemoteHangup_ = function() {
--- a/src/web_app/js/background.js
+++ b/src/web_app/js/background.js
@@ -46,20 +46,15 @@
       var ws = new WebSocket(message.wssUrl);
       port.wssPostUrl_ = message.wssPostUrl;
       ws.onopen = function() {
-        trace('RWS: onopen');
         sendWSEventMessageToWindow(port, Constants.WS_EVENT_ONOPEN);
       };
       ws.onerror = function() {
-        trace('RWS: onerror.');
         sendWSEventMessageToWindow(port, Constants.WS_EVENT_ONERROR);
       };
       ws.onclose = function(event) {
-        trace('RWS: onclose - code: ' + event.code + ' reason: ' +
-            event.reason);
         sendWSEventMessageToWindow(port, Constants.WS_EVENT_ONCLOSE, event);
       };
       ws.onmessage = function(event) {
-        trace('RWS: onmessage: ' + event.data);
         sendWSEventMessageToWindow(port, Constants.WS_EVENT_ONMESSAGE, event);
       };
       port.webSocket_ = ws;
@@ -69,7 +64,6 @@
         try {
           port.webSocket_.send(message.data);
         } catch (ex) {
-          trace('RWS: exception sending: ' + ex);
           sendWSEventMessageToWindow(port, Constants.WS_EVENT_SENDERROR, ex);
         }
       } else {
--- a/src/web_app/js/background.js
+++ b/src/web_app/js/background.js
@@ -9,14 +9,152 @@
 /* More information about these options at jshint.com/docs/options */
 
 // Variables defined in and used from chrome.
-/* globals chrome */
+/* globals chrome, sendAsyncUrlRequest, Constants, parseJSON */
 
 'use strict';
-chrome.app.runtime.onLaunched.addListener(function() {
-  chrome.app.window.create('appwindow.html', {
-    'width': 800,
-    'height': 600,
-    'left': 0,
-    'top': 0
+(function() {
+  chrome.app.runtime.onLaunched.addListener(function() {
+    chrome.app.window.create('appwindow.html', {
+      'width': 800,
+      'height': 600,
+      'left': 0,
+      'top': 0
+    });
   });
-});
+
+  // Send event notification from background to window.
+  var sendWSEventMessageToWindow = function(port, wsEvent, data) {
+    var message = {
+        action: Constants.WS_ACTION,
+        wsAction: Constants.EVENT_ACTION,
+        wsEvent: wsEvent
+    };
+    if (data) {
+      message.data = data;
+    }
+    trace('B -> W: ' + JSON.stringify(message));
+    try {
+      port.postMessage(message);
+    } catch (ex) {
+      trace('Error sending message: ' + ex);
+    }
+  };
+
+  var handleWebSocketRequest = function(port, message) {
+    if (message.wsAction === Constants.WS_CREATE_ACTION) {
+      trace('RWS: creating web socket: ' + message.wssUrl);
+      var ws = new WebSocket(message.wssUrl);
+      port.wssPostUrl_ = message.wssPostUrl;
+      ws.onopen = function() {
+        trace('RWS: onopen');
+        sendWSEventMessageToWindow(port, Constants.WS_EVENT_ONOPEN);
+      };
+      ws.onerror = function() {
+        trace('RWS: onerror.');
+        sendWSEventMessageToWindow(port, Constants.WS_EVENT_ONERROR);
+      };
+      ws.onclose = function(event) {
+        trace('RWS: onclose - code: ' + event.code + ' reason: ' +
+            event.reason);
+        sendWSEventMessageToWindow(port, Constants.WS_EVENT_ONCLOSE, event);
+      };
+      ws.onmessage = function(event) {
+        trace('RWS: onmessage: ' + event.data);
+        sendWSEventMessageToWindow(port, Constants.WS_EVENT_ONMESSAGE, event);
+      };
+      port.webSocket_ = ws;
+    } else if (message.wsAction === Constants.WS_SEND_ACTION) {
+      trace('RWS: sending: ' + message.data);
+      if (port.webSocket_ && port.webSocket_.readyState === WebSocket.OPEN) {
+        try {
+          port.webSocket_.send(message.data);
+        } catch (ex) {
+          trace('RWS: exception sending: ' + ex);
+          sendWSEventMessageToWindow(port, Constants.WS_EVENT_SENDERROR, ex);
+        }
+      } else {
+        // Attempt to send message using wss port url.
+        trace('RWS: web socket not ready, falling back to POST.');
+        var msg = parseJSON(message.data);
+        if (msg) {
+          sendAsyncUrlRequest('POST', port.wssPostUrl_, msg.msg);
+        }
+      }
+    } else if (message.wsAction === Constants.WS_CLOSE_ACTION) {
+      trace('RWS: close');
+      if (port.webSocket_) {
+        port.webSocket_.close();
+        port.webSocket = null;
+      }
+    }
+  };
+
+  var executeCleanupTask = function(port, message) {
+    trace('Executing queue action: ' + JSON.stringify(message));
+    if (message.action === Constants.XHR_ACTION) {
+      var method = message.method;
+      var url = message.url;
+      var body = message.body;
+      return sendAsyncUrlRequest(method, url, body);
+    } else if (message.action === Constants.WS_ACTION) {
+      handleWebSocketRequest(port, message);
+    } else {
+      trace('Unknown action in cleanup queue: ' + message.action);
+    }
+  };
+
+  var executeCleanupTasks = function(port, queue) {
+    var promise = Promise.resolve();
+    if (!queue) {
+      return promise;
+    }
+
+    var catchFunction = function(error) {
+      trace('Error executing cleanup action: ' + error.message);
+    };
+
+    while (queue.length > 0) {
+      var queueMessage = queue.shift();
+      promise = promise.then(
+          executeCleanupTask.bind(null, port, queueMessage)
+      ).catch(catchFunction);
+    }
+    return promise;
+  };
+
+  var handleMessageFromWindow = function(port, message) {
+    var action = message.action;
+    if (action === Constants.WS_ACTION) {
+      handleWebSocketRequest(port, message);
+    } else if (action === Constants.QUEUECLEAR_ACTION) {
+      port.queue_ = [];
+    } else if (action === Constants.QUEUEADD_ACTION) {
+      if (!port.queue_) {
+        port.queue_ = [];
+      }
+      port.queue_.push(message.queueMessage);
+    } else {
+      trace('Unknown action from window: ' + action);
+    }
+  };
+
+  chrome.runtime.onConnect.addListener(function(port) {
+    port.onDisconnect.addListener(function() {
+      // Execute the cleanup queue.
+      executeCleanupTasks(port, port.queue_).then(function() {
+        // Close web socket.
+        if (port.webSocket_) {
+          trace('Closing web socket.');
+          port.webSocket_.close();
+          port.webSocket_ = null;
+        }
+      });
+    });
+
+    port.onMessage.addListener(function(message) {
+      // Handle message from window to background.
+      trace('W -> B: ' + JSON.stringify(message));
+      handleMessageFromWindow(port, message);
+    });
+  });
+})();
--- a/src/web_app/js/background_test.js
+++ b/src/web_app/js/background_test.js
@@ -1,0 +1,248 @@
+/*
+ *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+ */
+
+/* More information about these options at jshint.com/docs/options */
+
+/* globals TestCase, assertEquals, xhrs:true, MockWebSocket, FAKE_WSS_POST_URL,
+   Constants, webSockets:true, FAKE_WSS_URL, WebSocket:true, MockXMLHttpRequest,
+   XMLHttpRequest:true, FAKE_SEND_EXCEPTION */
+
+'use strict';
+var FAKE_MESSAGE = JSON.stringify({
+        cmd: 'send',
+        msg: JSON.stringify({type: 'bye'})
+      });
+
+var MockEvent = function(addListener) {
+  this.addListener_ = addListener;
+};
+
+MockEvent.prototype.addListener = function(callback) {
+  this.addListener_(callback);
+};
+
+var MockPort = function() {
+  this.onDisconnectCallback_ = null;
+  this.onMessageCallback_ = null;
+  this.onPostMessage = null;
+
+  this.onDisconnect = new MockEvent(function(callback) {
+    this.onDisconnectCallback_ = callback;
+  }.bind(this));
+
+  this.onMessage = new MockEvent(function(callback) {
+    this.onMessageCallback_ = callback;
+  }.bind(this));
+};
+
+MockPort.prototype.disconnect = function() {
+  if (this.onDisconnectCallback_) {
+    this.onDisconnectCallback_();
+  }
+};
+
+MockPort.prototype.message = function(message) {
+  if (this.onMessageCallback_) {
+    this.onMessageCallback_(message);
+  }
+};
+
+MockPort.prototype.postMessage = function(message) {
+  if (this.onPostMessage) {
+    this.onPostMessage(message);
+  }
+};
+
+MockPort.prototype.createWebSocket = function() {
+  assertEquals(0, webSockets.length);
+
+  this.message({
+    action: Constants.WS_ACTION,
+    wsAction: Constants.WS_CREATE_ACTION,
+    wssUrl: FAKE_WSS_URL,
+    wssPostUrl: FAKE_WSS_POST_URL
+  });
+
+  assertEquals(1, webSockets.length);
+
+  assertEquals(WebSocket.CONNECTING, this.webSocket_.readyState);
+};
+
+var BackgroundTest = new TestCase('BackgroundTest');
+
+BackgroundTest.prototype.setUp = function() {
+  webSockets = [];
+  xhrs = [];
+
+  this.realWebSocket = WebSocket;
+  WebSocket = MockWebSocket;
+
+  this.mockPort_ = new MockPort();
+  window.chrome.callOnConnect(this.mockPort_);
+};
+
+BackgroundTest.prototype.tearDown = function() {
+  WebSocket = this.realWebSocket;
+};
+
+BackgroundTest.prototype.testCreateWebSocket = function() {
+  this.mockPort_.createWebSocket();
+};
+
+BackgroundTest.prototype.testCloseWebSocket = function() {
+  this.mockPort_.createWebSocket();
+
+  this.mockPort_.message({
+    action: Constants.WS_ACTION,
+    wsAction: Constants.WS_CLOSE_ACTION
+  });
+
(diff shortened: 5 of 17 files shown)


src/web_app/js/constants.js
```diff
@@ -0,0 +1,25 @@
...
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+ */
+
+/* More information about these options at jshint.com/docs/options */
+
+/* exported Constants */
+'use strict';
+
+var Constants = {
```
jiayliu: Please document what each constant is for.

src/web_app/js/background.js
```diff
@@ -9,14 +9,145 @@
...
+      ws.onmessage = function(event) {
+        trace('RWS: onmessage: ' + event.data);
+        sendMessageToWindow(port, {
+          action: Constants.WS_ACTION,
+          wsAction: Constants.ACTION_EVENT,
+          wsEvent: Constants.EVENT_ONMESSAGE,
+          data: event
+        });
+      };
+      port.webSocket_ = ws;
```
jiayliu: Encapsulate webSocket_ in Port and provide access methods?

src/web_app/js/port.js
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+/* More information about these options at jshint.com/docs/options */
+
+/* globals trace, chrome */
+/* exported Port */
+
+'use strict';
+
+var Port = (function() {
```
jiayliu: I thought this is used in background.js, but it appears to be used on the window side. Could you make the name clearer, e.g. WindowPort?

src/web_app/js/remotewebsocket.js
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+/* More information about these options at jshint.com/docs/options */
+
+/* globals Port, Constants */
+/* exported RemoteWebSocket */
+
+'use strict';
+
+var RemoteWebSocket = function(wssUrl) {
```
jiayliu: Please add comments for the purpose of the class.

chuckhays:
Updated with feedback.

I deleted SyncPromise like we talked about. After thinking it through, I think it should work fine with regular promises. We just have to be careful if/when we change the hangup method to be sure we don't add anything that cedes control of the thread before it is done.

Ignore the series of commits, I'll squash them before merging anything later.

src/web_app/js/windowport.js
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+/* More information about these options at jshint.com/docs/options */
+
+/* globals trace, chrome */
+/* exported WindowPort */
+
+'use strict';
+
+var WindowPort = (function() {
```
jiayliu: Please document the purpose of the object.

src/web_app/js/util.js
```diff
@@ -37,21 +37,44 @@ function queryStringToDictionary(queryString) {
 
 // Sends the URL request and returns a Promise as the result.
 function sendAsyncUrlRequest(method, url, body) {
+  return sendUrlRequest(method, url, true, body);
+}
+
+function sendSyncUrlRequest(method, url, body) {
```
jiayliu: This seems unused.

Gruntfile.js
```diff
@@ -133,14 +133,18 @@ module.exports = function(grunt) {
...
+            'src/web_app/js/constants.js',
             'src/web_app/js/infobox.js',
             'src/web_app/js/peerconnectionclient.js',
             'src/web_app/js/roomselection.js',
+            'src/web_app/js/remotewebsocket.js',
             'src/web_app/js/sdputils.js',
             'src/web_app/js/signalingchannel.js',
             'src/web_app/js/stats.js',
             'src/web_app/js/storage.js',
+            'src/web_app/js/syncpromise.js',
```
jiayliu: remove syncpromise.js

src/web_app/js/background.js
```diff
@@ -9,14 +9,149 @@
...
+          action: Constants.WS_ACTION,
+          wsAction: Constants.EVENT_ACTION,
+          wsEvent: Constants.WS_EVENT_ONMESSAGE,
+          data: event
+        });
+      };
+      port.webSocket_ = ws;
+    } else if (message.wsAction === Constants.WS_SEND_ACTION) {
+      trace('RWS: sending: ' + message.data);
+      port.webSocket_.send(message.data);
```
jiayliu: Calling WebSocket methods may throw exceptions and we need to handle.

src/web_app/js/background.js
```diff
@@ -9,14 +9,149 @@
...
+    } else if (message.wsAction === Constants.WS_CLOSE_ACTION) {
+      trace('RWS: close');
+      port.webSocket_.close();
+      port.webSocket = null;
+    }
+  };
+
+  var createQueuePromise = function(port, message) {
```
jiayliu: nit: rename to createCleanupTaskPromise?

src/web_app/js/appcontroller.js
```diff
@@ -193,7 +193,14 @@ AppController.prototype.showRoomSelection_ = function() {
...
   this.call_.start(roomId);
 
-  window.onbeforeunload = this.call_.hangup.bind(this.call_);
+  // Chrome apps can't use onbeforeunload.
+  if (!isChromeApp()) {
+    // Call hangup with async = false. Required to complete multiple
+    // clean up steps before page is closed.
+    window.onbeforeunload = function() {
+      this.call_.hangup(false);
+    }.bind(this);
```
milkypostman: I think this can be:

```javascript
window.onbeforeunload = this.call_.hangup.bind(this, false);
```

src/web_app/js/background.js
```diff
@@ -9,14 +9,149 @@
...
+    };
+  };
+
+  var createQueueCatch = function() {
+    return function(error) {
+      trace('Error executing cleanup action: ' + error.message);
+    };
+  };
+
+  var executeQueue = function(port, queue) {
```
jiayliu: nit: rename executeCleanupTasks?

src/web_app/js/appcontroller.js
```diff
@@ -193,7 +193,14 @@ AppController.prototype.showRoomSelection_ = function() {
...
+  }
```
milkypostman: Can this change be combined with the other `if` statement below so there is just a single code path where we are considering non chrome app?

src/web_app/js/background.js
```diff
@@ -9,14 +9,149 @@
...
+      } else if (message.action === Constants.WS_ACTION) {
+        handleWebSocketRequest(port, message);
+      } else {
+        trace('Unknown action in cleanup queue: ' + message.action);
+      }
+    };
+  };
+
+  var createQueueCatch = function() {
+    return function(error) {
```
jiayliu: nit: inline the method?

src/web_app/js/background.js
```diff
@@ -9,14 +9,149 @@
...
+      trace('Unknown action from window: ' + action);
+    }
+  };
+
+  chrome.runtime.onConnect.addListener(function(port) {
+    port.onDisconnect.addListener(function() {
+      // Execute the cleanup queue.
+      executeQueue(port, port.queue_).then(function() {
+        // Close web socket.
+        if (port.webSocket_) {
```
jiayliu: Check if the websocket readystate is open?

src/web_app/js/background.js
jiayliu: And need to check websocket state.

src/web_app/js/background_test.js
```diff
@@ -0,0 +1,195 @@
...
+  this.mockPort_.createWebSocket();
+  var message = null;
+  this.mockPort_.onPostMessage = function(msg) {
+    message = msg;
+  };
+
+  var ws = this.mockPort_.webSocket_;
+
+  ws.onopen();
+  assertEquals(Constants.WS_ACTION, message.action);
```
jiayliu: Consolidate checking action/wsAction/wsevent into a helper method?

src/web_app/js/call.js
```diff
@@ -89,19 +133,42 @@ Call.prototype.hangup = function() {
...
-  this.channel_.send(JSON.stringify({type: 'bye'}));
-  this.channel_.close();
+  // This section of code is executed in both sync and async depending on
+  // where it is called from. When the browser is closed, the requests must
+  // be executed as sync to finish before the browser closes. When called
+  // from pressing the hang up button, the requests are executed async.
+  var path = this.createLeaveUrl_();
+
+  var sendLeave = sendUrlRequest('POST', path, async).catch(function(error) {
+    trace('Error sending /leave: ' + error.message);
```
jiayliu: nit: create helper method to trace error.message?

src/web_app/js/call.js
```diff
@@ -89,19 +133,42 @@ Call.prototype.hangup = function() {
...
+  }.bind(this));
+
+  var sendBye = sendLeave.then(function() {
+    // Send bye to other client.
```
jiayliu: nit: to the other client.

src/web_app/js/common_test.js
```diff
@@ -0,0 +1,90 @@
...
+
+/* More information about these options at jshint.com/docs/options */
+
+/* globals assertEquals */
+/* exported FAKE_WSS_POST_URL, FAKE_WSS_URL, FAKE_WSS_POST_URL, FAKE_ROOM_ID,
+   FAKE_CLIENT_ID, MockWebSocket, MockXMLHttpRequest, webSockets, xhrs */
+
+'use strict';
+
+var FAKE_WSS_URL = 'wss://foo.com';
```
jiayliu: nit: rename the file to test_mocks.js?

src/web_app/js/background.js
```diff
@@ -9,14 +9,149 @@
...
   });
-});
+
+  // Send message from background to window.
+  var sendMessageToWindow = function(port, message) {
+    trace('B -> W: ' + JSON.stringify(message));
+    try {
+      port.postMessage(message);
+    }
+    catch (ex) {
```
jiayliu: nit: catch on the same line as }

src/web_app/js/background.js
```diff
@@ -9,14 +9,149 @@
...
+        trace('RWS: onopen');
+        sendMessageToWindow(port, {
+          action: Constants.WS_ACTION,
+          wsAction: Constants.EVENT_ACTION,
+          wsEvent: Constants.WS_EVENT_ONOPEN
+        });
+      };
+      ws.onerror = function() {
+        trace('RWS: onerror.');
+        sendMessageToWindow(port, {
```
jiayliu: nit: make sendMessageToWindow take arguments |action|, |wsAction|, |wsEvent| and |data|?

src/web_app/js/call.js
```diff
@@ -61,6 +61,45 @@ Call.prototype.start = function(roomId) {
   }
 };
 
+Call.prototype.createCleanupQueue_ = function() {
```
jiayliu: Add test to verify cleanup tasks are queued and removed correctly?

src/web_app/js/appcontroller.js
chuckhays: I had it that was, but it was causing odd side effects, possibly because onbeforeunload calls the function with an event object. For some reason it was causing one of those "Are you sure you want to leave this page?" pop ups when you closed the page.

milkypostman: ahh, yes.  because the function is returning something. non-null return value produces a confirmation prompt.

src/web_app/js/util.js
chuckhays: It is unused. It was just there as a convenience method. Removed it.

src/web_app/js/appcontroller.js
chuckhays: I suspected that, but never verified.

src/web_app/js/background.js
chuckhays: One of our jslint/jscs tools doesn't allow creating methods in a loop, to avoid modified closures. It isn't smart enough to detect that you aren't using any variables from the closure. This is the simplest way to make the rule happy. The other option is to disable the rule for that block.

jiayliu: Cache the function in a local var so we don't create a new closure?

chuckhays:
Made most of the changes, working on a unit test for the queue creation, and I still need to do the error handling for the web socket in background.js.

src/web_app/js/background.js
chuckhays: I added the POST to wss path like signalingchannel has. We shouldn't get into that state because the readystate is mirrored on the window side, but good to have the code there in case it happens.

chuckhays:
I think I've addressed the comments, just need to decide what we want to do with any exceptions that .send() might throw.

src/web_app/js/background.js
```diff
@@ -9,14 +9,151 @@
...
+        }
+      }
+    } else if (message.wsAction === Constants.WS_CLOSE_ACTION) {
+      trace('RWS: close');
+      port.webSocket_.close();
+      port.webSocket = null;
+    }
+  };
+
+  var createCleanupTaskPromise = function(port, message) {
```
jiayliu: If you call the inner function executeCleanupTask(port, message), you can just make the caller do 
promise.then(executeCleanupTask.bind(port, message))

chuckhays:
Do either of you have an opinion on what to do with exceptions thrown by .send in the websocket owned by background.js? @tkchin @jiayliu

src/web_app/js/background.js
```diff
@@ -9,14 +9,151 @@
...
+        // Attempt to send message using wss port url.
+        trace('RWS: web socket not ready, falling back to POST.');
+        var msg = parseJSON(message.data);
+        if (msg) {
+          sendAsyncUrlRequest('POST', port.wssPostUrl_, msg.msg);
+        }
+      }
+    } else if (message.wsAction === Constants.WS_CLOSE_ACTION) {
+      trace('RWS: close');
+      port.webSocket_.close();
```
jiayliu: check webSocket_ is not null?

src/web_app/js/background.js
```diff
@@ -9,14 +9,151 @@
...
+        sendWSEventMessageToWindow(port, Constants.WS_EVENT_ONMESSAGE, event);
+      };
+      port.webSocket_ = ws;
+    } else if (message.wsAction === Constants.WS_SEND_ACTION) {
+      trace('RWS: sending: ' + message.data);
+      if (port.webSocket_ && port.webSocket_.readyState === WebSocket.OPEN) {
+        try {
+          port.webSocket_.send(message.data);
+        } catch (ex) {
+          trace('RWS: exception sending: ' + ex);
```
jiayliu: We probably want to populate the error back to the window

jiayliu:
I think we need to send the error to the window, so we can update UI, etc

On Fri, Feb 20, 2015 at 2:42 PM, Chuck Hays <notifications@github.com>
wrote:

> Do either of you have an opinion on what to do with exceptions thrown by
> .send in the websocket owned by background.js? @tkchin
> <https://github.com/tkchin> @jiayliu <https://github.com/jiayliu>
>
> —
> <https://github.com/webrtc/apprtc/pull/82#issuecomment-75333168>.
>



-- 
Jiayang Liu

src/web_app/js/background.js
chuckhays: Would that have to be .bind(null, port, message) in this case?

src/web_app/js/call.js
```diff
@@ -89,19 +131,45 @@ Call.prototype.hangup = function() {
...
+  }.bind(this)).catch(function(error) {
+    trace('Error closing signaling channel: ' + error.message);
+  }.bind(this));
+
+  var setParams = closeChannel.then(function() {
+    this.params_.previousRoomId = this.params_.roomId;
+    this.params_.roomId = null;
+    this.params_.clientId = null;
+  }.bind(this));
+
```
jiayliu: For sync mode, check if everything is done and spew error if not?

src/web_app/js/background.js
chuckhays: I think that makes sense, right now we don't handle this type of exception in signalingchannel, but we should probably do the same thing in both places.

src/web_app/js/call.js
chuckhays: Could you clarify what you mean? The sync mode only runs when the tab/window is closed.

src/web_app/js/call_test.js
```diff
@@ -58,3 +65,36 @@ CallTest.prototype.testRestartInitializesMedia = function() {
   call.restart();
   assertTrue(mediaStarted);
 };
+
+CallTest.prototype.testSetUpCleanupQueue = function() {
```
jiayliu: Add test for hangup?

src/web_app/js/remotewebsocket.js
```diff
@@ -0,0 +1,76 @@
...
+RemoteWebSocket.prototype.send = function(data) {
+  this.sendMessage_({
+    action: Constants.WS_ACTION,
+    wsAction: Constants.WS_SEND_ACTION,
+    data: data
+  });
+};
+
+RemoteWebSocket.prototype.close = function() {
+  this.readyState = WebSocket.CLOSING;
```
jiayliu: Should check readyState is not closed or closing

src/web_app/js/remotewebsocket.js
```diff
@@ -0,0 +1,76 @@
...
+  });
+  this.readyState = WebSocket.CONNECTING;
+};
+
+RemoteWebSocket.prototype.sendMessage_ = function(message) {
+  WindowPort.sendMessage(message);
+};
+
+RemoteWebSocket.prototype.send = function(data) {
+  this.sendMessage_({
```
jiayliu: should check readystate is open

src/web_app/js/call.js
jiayliu: Yes, it's still helpful to log it

src/web_app/js/background.js
jiayliu: Right

src/web_app/js/background.js
jiayliu: On Fri, Feb 20, 2015 at 2:51 PM, Chuck Hays <notifications@github.com>
wrote:

> In src/web_app/js/background.js
> <https://github.com/webrtc/apprtc/pull/82#discussion_r25110228>:
>
> > +        trace('RWS: onclose - code: ' + event.code + ' reason: ' +
> > +            event.reason);
> > +        sendWSEventMessageToWindow(port, Constants.WS_EVENT_ONCLOSE, event);
> > +      };
> > +      ws.onmessage = function(event) {
> > +        trace('RWS: onmessage: ' + event.data);
> > +        sendWSEventMessageToWindow(port, Constants.WS_EVENT_ONMESSAGE, event);
> > +      };
> > +      port.webSocket_ = ws;
> > +    } else if (message.wsAction === Constants.WS_SEND_ACTION) {
> > +      trace('RWS: sending: ' + message.data);
> > +      if (port.webSocket_ && port.webSocket_.readyState === WebSocket.OPEN) {
> > +        try {
> > +          port.webSocket_.send(message.data);
> > +        } catch (ex) {
> > +          trace('RWS: exception sending: ' + ex);
>
> I think that makes sense, right now we don't handle this type of exception
> in signalingchannel, but we should probably do the same thing in both
> places.
>
> I'm fine with doing it later if we didn't handle it before your change,
with a bug to track it.


> —
> <https://github.com/webrtc/apprtc/pull/82/files#r25110228>.
>



-- 
Jiayang Liu

src/web_app/js/call.js
chuckhays: I didn't understand what you mean here, what are you suggesting we log, and what conditions?

jiayliu: if (!async) {
  if (this.params.roomId != null || this.params.clientId != null) {
    trace("ERROR: Room state should be reset for sync hangup")
  }
}

src/web_app/chrome_app/manifest.json
```diff
@@ -5,7 +5,7 @@
   "manifest_version": 2,
   "app": {
     "background": {
-      "scripts": ["js/background.js"]
+      "scripts": ["js/apprtc.debug.js", "js/background.js"]
```
tkchin: Need to look at rest of CL first, but I think it'd be better to have a separate compile target for Chrome build with only the required bits in the background. E.g. apprtc.chrome.debug.js

src/web_app/js/call.js
```diff
@@ -89,19 +135,51 @@ Call.prototype.hangup = function() {
...
+  if (!async) {
+    if (this.params_.roomId !== null || this.params_.clientId !== null) {
+      trace('ERROR: sync cleanup tasks did not complete successfully.');
+    }
+  }
+
+  return setParams;
+};
+
+Call.prototype.createLeaveUrl_ = function() {
```
tkchin: nit: getLeaveUrl_

src/web_app/js/signalingchannel.js
```diff
@@ -122,14 +126,15 @@ SignalingChannel.prototype.close = function() {
...
-  xhr.open('DELETE', path, false);
-  xhr.send();
-
-  this.clientId_ = null;
-  this.roomId_ = null;
-  this.registered_ = false;
+  var path = this.getWssPostUrl();
+
+  return sendUrlRequest('DELETE', path, async).catch(function(error) {
+    trace('Error deleting web socket connection: ' + error.message);
```
tkchin: Not referencing this in the function, why bind?

src/web_app/js/util.js
```diff
@@ -37,21 +37,40 @@ function queryStringToDictionary(queryString) {
...
-            Error('Status=' + xhr.status + ', response=' + xhr.responseText));
+            Error('Status=' + xhr.status + ', response=' +
+                  xhr.responseText));
         return;
       }
       resolve(xhr.responseText);
     };
-    xhr.open(method, url, true);
+
+    var xhr;
```
tkchin: nit: declare this before reportResults

src/web_app/js/util.js
```diff
@@ -37,21 +37,40 @@ function queryStringToDictionary(queryString) {
...
+    xhr = new XMLHttpRequest();
+    if (async) {
+      xhr.onreadystatechange = function() {
```
tkchin: onreadystatechange isn't called for sync requests?

src/web_app/js/windowport.js
```diff
@@ -0,0 +1,49 @@
...
+/* globals trace, chrome */
+/* exported WindowPort */
+
+'use strict';
+
+// This is used to communicate from the Chrome App window to background.js.
+// It opens a Port object to send and receive messages. When the Chrome
+// App window is closed, background.js receives notification and can
+// handle clean up tasks.
+var WindowPort = (function() {
```
tkchin: Unusual (to my eyes at least, but not incorrect). Is there reason not to use the common way of defining new object? Why stuff it in a closure instead? If the goal is to have just one port object can wrap just the port in a closure. It looks like you're trying to add namespace here, but we haven't done that elsewhere in the project.

ie. 

var sendPortMessage;
var addPortMessageListener;
(function() {
  var port = chrome.runtime.connect();
  if (!port) { <report error> }
  sendPortMessage = function(message) {...};
  addPortMessageListener = function(listener) {...};
})();

or if multiple calls to port return the same channel:

var WindowPort = function() {
  this.port_ = chrome.runtime.connect();
};

WindowPort.prototype.sendMessage = function(message) ...
etc

src/web_app/js/call.js
```diff
@@ -209,6 +287,9 @@ Call.prototype.connectToRoom_ = function(roomId) {
     Promise.all([this.getTurnServersPromise_, this.getMediaPromise_])
         .then(function() {
           this.startSignaling_();
+          if (isChromeApp()) {
+            this.createCleanupQueue_();
```
tkchin: nit: queueCleanupMessages instead, not actually creating something here.
Can you add a comment explaining that we need to queue messages earlier because when window dies only background.js knows?

tkchin:
Just minor comments. I think this approach makes sense given our current priorities.

jiayliu:
lgtm barring Zeke's comments
Thanks!

src/web_app/js/signalingchannel.js
milkypostman: We never know what the future holds.

src/web_app/chrome_app/manifest.json
chuckhays: I thought about doing this, but we would be shipping the same code twice in the chrome app package, because appwindow uses apprtc.debug.js, because it uses the same code as the apprtc on the web.

src/web_app/js/signalingchannel.js
chuckhays: Habit mostly.

src/web_app/js/util.js
chuckhays: Interesting, I had thought it wasn't, but it appears that at least in Chrome and FF it is called even with async=false.

src/web_app/js/windowport.js
chuckhays: I was indeed trying to add a namespace, as we have so many globals already, it seems desirable to at least slow the rate of increase. I'll change it if you want though.

The difference between what I wrote and your first example is only the namespace (as lazy initialization of the port object) as far as I can tell.

For the second example, the actual port object (returned by connect()) should be a singleton. The reason is that on the background side, the port object is what ties the requests to their correct web socket, so the cleanup queue requests need to come to the same port to be able to use the web socket created by RemoteWebSocket's requests.

chuckhays:
@tkchin Your comment about xhr helped me discover that the sync cleanup wasn't actually running completely sync. Even if the promise has completed, .then apparently schedules the body to run on an immediate timeout or something similar. I had to change to another strategy @jiayliu and I had discussed to execute the cleanup steps to ensure they happen synchronously.

PTAL at the most recent change. I'll have to merge after, but there shouldn't be any major conflicts, last time I checked it was just in gruntfile.js.

src/web_app/js/windowport.js
tkchin: For namespace can we follow this pattern? (I think it's cleaner than repeating declaration at the end of the file. Also see style guide for namespace naming conventions).

var apprtc = apprtc || {};
apprtc.port = apprtc.port || {};
(function() {
  var port_ = chrome.runtime.connect();
  apprtc.port.sendMessage = function(message) {...}
  apprtc.port.addListener = function(listener) {...}
})();

src/web_app/chrome_app/manifest.json
tkchin: Well. As opposed to loading the same code twice? Because now everything is loaded in two pages when not all of it has to be. Seems to increase memory footprint unnecessarily. Don't have to fix it now, because I think either way package size / memory size is small. But it just felt the same like having a bunch of header files in an impl file that aren't actually being used, especially since they all pollute the global namespace currently.

tkchin:
Ok, so I'm guessing that when Promise then block executes async the page is gone since we're in the onbeforeunload handler? [Because promise chaining should be fine, the only sync requirement is that GAE receives /leave before we DELETE off Collider]

src/web_app/chrome_app/manifest.json
chuckhays: I'll add an issue to do this. Probably worth it to get things set up correctly sooner vs later.

chuckhays:
I squashed down most of the previous commits (leaving the last couple for now, will squash before merging) and rebased on master.

src/web_app/js/background.js
```diff
@@ -9,14 +9,152 @@
...
+    }
+  };
+
+  var handleWebSocketRequest = function(port, message) {
+    if (message.wsAction === Constants.WS_CREATE_ACTION) {
+      trace('RWS: creating web socket: ' + message.wssUrl);
+      var ws = new WebSocket(message.wssUrl);
+      port.wssPostUrl_ = message.wssPostUrl;
+      ws.onopen = function() {
+        trace('RWS: onopen');
```
jiayliu: nit: the trace before sendWSEventMessageToWindow could probably be removed, since sendWSEventMessageToWindow already calls trace.

jiayliu:
Could you add tests for Call.hangup()?

On Wed, Feb 25, 2015 at 3:41 PM, Chuck Hays <notifications@github.com>
wrote:

> I squashed down most of the previous commits (leaving the last couple for
> now, will squash before merging) and rebased on master.
>
> —
> <https://github.com/webrtc/apprtc/pull/82#issuecomment-76086355>.
>



-- 
Jiayang Liu

chuckhays:
Added unit test and removed verbose logging.

chuckhays:
The other problem with testing the async verison is:
https://github.com/webrtc/apprtc/issues/59

Because we use a chain of .then() in the code, the simplified promise implementation we use in the unit test causes only the last .then body to be used.

When I filed that bug, I tried to do a simple implementation to support chaining, but it got too complex, it was basically turning into a full fledged promises polyfill.

jiayliu:
Thanks for investigatin. I'm fine to leave it for now.
lgtm

On Fri, Feb 27, 2015 at 1:36 PM, Chuck Hays <notifications@github.com>
wrote:

> The other problem with testing the async verison is:
> #59 <https://github.com/webrtc/apprtc/issues/59>
>
> Because we use a chain of .then() in the code, the simplified promise
> implementation we use in the unit test causes only the last .then body to
> be used.
>
> When I filed that bug, I tried to do a simple implementation to support
> chaining, but it got too complex, it was basically turning into a full
> fledged promises polyfill.
>
> —
> <https://github.com/webrtc/apprtc/pull/82#issuecomment-76476298>.
>



-- 
Jiayang Liu

PR #84, opened 2015-02-19
Encrypt ndb and memcache entries.

tkchin:
@jiayliu @chuckhays PTAL

Had to grep for function invocations, I think I got all of them but if I didn't this may cause breakage. Tested locally with Open rooms. Could not test Direct rooms because GAE console seems to be down and I can't retrieve API key. Didn't test on live server because of shared ndb across versions. (Perhaps we should namespace that as well? Version can be retrieved from os.environ iirc).

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,20 +81,30 @@ When the `Analytics` class detects that AppRTC is running locally, all data is l
 
 ## GCM
 
-Call notifications are sent and received using GCM. In order for the server to send GCM messages, an API key is required. This key should be obtained from the developer console for the app engine project. In order to specify a key for the build, run,
+Call notifications are sent and received using GCM. In order for the server to send GCM messages, an API key is required. This key should be obtained from the developer console for the app engine project. It should be placed in a file named gcm_api_key in the following location:
 
 ```
-grunt build:<API_KEY>
+out/app_engine/gcm_api_key
 ```
 
-When building the app engine package only, run,
+This must be provided for GCM notifications to work correctly. See internal documentation (Google only) for a script that will generate this for you.
+
+## PyCrypto
+
+We use the PyCrypto module for AES encryption and SHA256 hashing. This module is not included in the Google AppEngine SDK. In order to run a local development server, please install PyCrypto into your local site-packages directory. You can download it from:
+
+https://pypi.python.org/pypi/pycrypto
+
+If using pip or easy_install you may run into egg problems with the dev appserver; if so, install the module manually.
+
+Two files should be provided, one named aes_key and another named hash_salt. They should contain the relevant encryption key and hash salt and be placed in:
 
 ```
-grunt shell:buildAppEnginePackage:<API_KEY>
+out/app_engine/aes_key
+out/app_engine/hash_salt
 ```
 
-This must be given for GCM notifications to work correctly.
-
+These must be provided for encryption to work correctly. See internal documentation (Google only) for a script that will generate this for you.
 
 ## BigQuery
 
--- a/build/build_app_engine_package.py
+++ b/build/build_app_engine_package.py
@@ -58,11 +58,6 @@ def build_version_info_file(dest_path):
   except IOError as e:
     print str(e)
 
-def write_gcm_config_json(out_dir, gcm_api_key):
-  path = os.path.join(out_dir, 'gcm_config.json')
-  with open(path, 'w') as f:
-    f.write(json.dumps({'GCM_API_KEY': gcm_api_key}))
-
 def CopyApprtcSource(src_path, dest_path):
   if os.path.exists(dest_path):
     shutil.rmtree(dest_path)
@@ -101,26 +96,19 @@ def CopyApprtcSource(src_path, dest_path):
 
   build_version_info_file(os.path.join(dest_path, 'version_info.json'))
 
-
 def main():
   parser = optparse.OptionParser(USAGE)
   parser.add_option("-t", "--include-tests", action="store_true",
                     help='Also copy python tests to the out dir.')
   options, args = parser.parse_args()
-  if len(args) < 2:
-    parser.error('At least 2 arguments required.')
+  if len(args) != 2:
+    parser.error('Exactly 2 arguments required.')
 
   src_path, dest_path = args[0:2]
   CopyApprtcSource(src_path, dest_path)
   if options.include_tests:
     app_engine_code = os.path.join(src_path, 'app_engine')
     test_file_herder.CopyTests(os.path.join(src_path, 'app_engine'), dest_path)
-
-  gcm_api_key = None
-  if len(args) > 2:
-    gcm_api_key = args[2]
-    write_gcm_config_json(dest_path, gcm_api_key)
-
 
 if __name__ == '__main__':
   sys.exit(main())
--- a/src/app_engine/encryptor.py
+++ b/src/app_engine/encryptor.py
@@ -1,0 +1,176 @@
+"""AES Encryption module.
+"""
+
+import logging
+import struct
+
+from Crypto import Random
+from Crypto.Cipher import AES
+from Crypto.Hash import SHA256
+
+import constants
+import util
+
+AES_KEY = None
+AES_KEY_PATH = 'gae_aes_key'
+HASH_SALT = None
+HASH_SALT_PATH = 'gae_hash_salt'
+
+
+def get_aes_key():
+  """Reads the AES key from disk."""
+  global AES_KEY
+  if AES_KEY is not None:
+    return AES_KEY
+  if not util.file_exists(AES_KEY_PATH):
+    if not constants.IS_DEV_SERVER:
+      logging.error('Missing AES key file.')
+    return None
+  key = util.read_file_contents(AES_KEY_PATH)
+  if key is None or len(key) not in AES.key_size:
+    logging.error('Failed to load AES key.')
+  else:
+    AES_KEY = key
+  return AES_KEY
+
+
+def get_hash_salt():
+  """Reads the hash salt from disk."""
+  global HASH_SALT
+  if HASH_SALT is not None:
+    return HASH_SALT
+  if not util.file_exists(HASH_SALT_PATH):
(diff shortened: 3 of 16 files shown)


README.md
```diff
@@ -81,20 +81,30 @@ When the `Analytics` class detects that AppRTC is running locally, all data is l
...
 
 ```
-grunt shell:buildAppEnginePackage:<API_KEY>
+out/app_engine/aes_key
+out/app_engine/hash_salt
 ```
 
-This must be given for GCM notifications to work correctly.
-
+These must be provided for encryption to work correctly. See internal documentation for a script that will generate this for you.
```
jiayliu: Make it clear that the "internal documentation" applies to googlers.

build/build_app_engine_package.py
```diff
@@ -56,12 +56,8 @@ def build_version_info_file(dest_path):
   except IOError as e:
     print str(e)
 
-def write_gcm_config_json(out_dir, gcm_api_key):
-  path = os.path.join(out_dir, 'gcm_config.json')
-  with open(path, 'w') as f:
-    f.write(json.dumps({'GCM_API_KEY': gcm_api_key}))
 
-def main(src_path, dest_path, gcm_api_key=None):
+def main(src_path, dest_path):
```
jiayliu: Merge with the changes from Master?

src/app_engine/gcmrecord.py
```diff
@@ -1,59 +1,159 @@
...
-  gcm_id = ndb.StringProperty()
-  code = ndb.StringProperty()
-  code_sent_time = ndb.DateTimeProperty()
-  verified = ndb.BooleanProperty()
-  last_modified_time = ndb.DateTimeProperty()
+class GCMEncryptedRecord(ndb.Model):
+  """Stores an encrypted registration in db."""
+  user_key = ndb.BlobProperty(indexed=True)
+  gcm_key = ndb.BlobProperty(indexed=True)
+  content = ndb.BlobProperty()
```
jiayliu: Please comment that content is the encrypted json representation of GCMRecord, user_key is the hash of user id, gcm_key is the hash of gcm id.

src/app_engine/gcmrecord.py
```diff
@@ -1,59 +1,159 @@
...
 
-class GCMRecord(ndb.Model):
-  """Models a user ID to GCM ID binding record."""
-  user_id = ndb.StringProperty()
-  gcm_id = ndb.StringProperty()
-  code = ndb.StringProperty()
-  code_sent_time = ndb.DateTimeProperty()
-  verified = ndb.BooleanProperty()
-  last_modified_time = ndb.DateTimeProperty()
+class GCMEncryptedRecord(ndb.Model):
```
jiayliu: Tests for this?

src/app_engine/room.py
```diff
@@ -111,14 +156,21 @@ def get_memcache_key_for_room(host, room_id):
...
-  return room is not None
+  encrypted_room = memcache_client.get(key)
+  return encrypted_room is not None
 
 
 def get_room(host, room_id):
   key = get_memcache_key_for_room(host, room_id)
   memcache_client = memcache.Client()
-  return memcache_client.get(key)
+  encrypted_room = memcache_client.get(key)
```
jiayliu: Tests for room.py?

README.md
jiayliu: Can one run the server locally in no-encryption mode if these files are missing?

jiayliu: How long does encryption/decryption take?

jiayliu:
Nice work! Looks pretty good in general.

src/app_engine/encryptor.py
```diff
@@ -0,0 +1,170 @@
...
+from Crypto.Cipher import AES
+from Crypto.Hash import SHA256
+
+AES_KEY = None
+AES_KEY_PATH = 'aes_key'
+HASH_SALT = None
+HASH_SALT_PATH = 'hash_salt'
+
+
+def read_file_contents(file_name):
```
chuckhays: This looks like it could/should be used for reading the gcm file as well?

src/app_engine/encryptor.py
```diff
@@ -0,0 +1,170 @@
...
+    Padding is required because AES required 16-byte aligned content.
+
+    Args:
+      message: A message string.
+
+    Returns:
+      An encrypted and signed byte string.
+    """
+    key = get_aes_key()
+    if key is None:
```
chuckhays: I'm not sure what the right thing to do here is, but returning the message back in plaintext seems dangerous. Maybe we should only allow this mode with a flag? It seems safer to refuse to run in the default case if the key is missing.

src/app_engine/encryptor.py
```diff
@@ -0,0 +1,170 @@
...
+    return message
+
+
+class Hasher(object):
+   """Convenience class to generate SHA256 hashes using a global salt."""
+
+   @classmethod
+   def salted_hash(cls, message):
+     salt = get_hash_salt()
+     if salt is None:
```
chuckhays: Same comment here as with the key.

src/app_engine/encryptor_test.py
```diff
@@ -0,0 +1,56 @@
...
+
+    # Then activate the testbed, which prepares the service stubs for use.
+    self.testbed.activate()
+
+    # Override AES key for testing.
+    encryptor.AES_KEY = 'a' * 16
+    encryptor.HASH_SALT = 'a' * 16
+
+  def testEncryptAndDecrypt(self):
+    message = b'All your base are belong to us!'
```
chuckhays: With the padding and unpadding code, it might make sense to test this with a couple of different length strings.

chuckhays: Curiousity question, what is the purpose of the b prefix here?

src/app_engine/gcmrecord.py
```diff
@@ -1,59 +1,159 @@
...
+  def parse_encrypted(cls, encrypted_content):
+    """Inflates a GCMRecord from previously generated encrypted content."""
+    content = Encryptor.decrypt(encrypted_content)
+    try:
+      content_dictionary = json.loads(content)
+    except ValueError as e:
+      logging.error('Bad encrypted content. Error:%s', str(e))
+      return None
+    record = GCMRecord()
+    record.user_id = content_dictionary.get(cls.USER_ID_KEY)
```
chuckhays: Do we want to reject the record if one of the fields is missing?

src/app_engine/gcmrecord.py
```diff
@@ -61,94 +161,118 @@ def get_associated_records_for_gcm_id(cls, gcm_id, verified_only = False):
...
+    """Marks the registration as verified if the supplied code matches."""
+    gcm_key = Hasher.salted_hash(gcm_id)
+    encrypted_records = GCMEncryptedRecord.get_by_gcm_key(gcm_key)
+    assert len(encrypted_records) < 2
+    if not encrypted_records:
+      return constants.RESPONSE_NOT_FOUND
+    encrypted_record = encrypted_records[0]
+    record = GCMRecord.parse_encrypted(encrypted_record.content)
+    if not record:
+      encrypted_record.key.delete()
```
chuckhays: What does this do?

src/app_engine/gcmrecord.py
```diff
@@ -61,94 +161,118 @@ def get_associated_records_for_gcm_id(cls, gcm_id, verified_only = False):
...
-        return constants.RESPONSE_SUCCESS
-      else:
-        logging.warning('Cannot update unverified GCM binding, ' \
-          'user_id=%s, old_gcm_id=%s, new_gcm_id=%s' \
-           % (user_id, old_gcm_id, new_gcm_id))
-        return constants.RESPONSE_INVALID_STATE
-    return constants.RESPONSE_NOT_FOUND
+    """Updates the registration with a new GCM id."""
+    old_gcm_key = Hasher.salted_hash(old_gcm_id)
+    encrypted_records = GCMEncryptedRecord.get_by_gcm_key(old_gcm_key)
```
chuckhays: It looks like before we checked user_id on the record as well, do we care?

chuckhays:
I think this is looking good, just made a couple of comments/questions.

README.md
tkchin: Done / Yes.
On local machine roughly 0.088ms encrypt 0.029ms decrypt for a 1000-character bytestring. See unit test.

src/app_engine/encryptor.py
tkchin: Yeah. Made it so that plaintext only if local dev server.

src/app_engine/encryptor_test.py
tkchin: To indicate byte string.

src/app_engine/gcmrecord.py
tkchin: Done.

src/app_engine/gcmrecord.py
tkchin: On second thought, let's not delete so we can potentially see what happened by examining the store.

src/app_engine/gcmrecord.py
tkchin: GCM ID is unique so there shouldn't be a need to check user id.

tkchin:
Rebased off merge. PTAL.
(I'll flatten this after LGTMs)

src/app_engine/encryptor_test.py
chuckhays: Sorry, I meant why use it here? The docs are a bit confusing, but it appears to do nothing for python 2.x, and in 3 it will make this string not the same as str, which is what json.dumps() returns.

tkchin: Ah! Yes, not needed. I'd initially planned for encryptor to take in byte strings only.
Yeah it doesn't actually do anything, but good to keep track of what is to be intepreted as bytes, and what can be intepreted as UTF8.

chuckhays:
LGTM, you still have a couple of b' prefixed strings in the changes, not sure if that was intentional or not.

jiayliu:
lgtm

PR #85, opened 2015-02-20
Add a LogManager for Collider which creates a separate log file on each ...

jiayliu:
...day and deletes logs older than 30 days.

### DIFF
--- a/src/collider/collider/logmanager.go
+++ b/src/collider/collider/logmanager.go
@@ -1,0 +1,143 @@
+// Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file in the root of the source
+// tree.
+
+// Package collider implements a signaling server based on WebSocket.
+package collider
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"path"
+	"strings"
+	"time"
+)
+
+// LogManager redirects log output to files, one for each |cycle|, and deletes files that are last modified before the retention period.
+type LogManager struct {
+	dir       string
+	cycle     time.Duration
+	retention time.Duration
+	file      *os.File
+	stopped   chan bool
+	ticker    *time.Ticker
+}
+
+func NewLogManager(d string, c time.Duration, r time.Duration) *LogManager {
+	if d == "" {
+		log.Printf("Invalid log dir")
+		return nil
+	}
+	if c >= r {
+		log.Printf("Log retention time must be greater than log cycle")
+		return nil
+	}
+	return &LogManager{
+		dir:       d,
+		cycle:     c,
+		retention: r,
+	}
+}
+
+func (lm *LogManager) Start() error {
+	if lm.stopped != nil || lm.ticker != nil {
+		log.Printf("Start called in invalid state")
+		return errors.New("Invalid state")
+	}
+
+	lm.stopped = make(chan bool)
+	lm.ticker = time.NewTicker(lm.cycle)
+	go lm.runTicker()
+
+	lm.startNewLog()
+	return nil
+}
+
+func (lm *LogManager) Stop() {
+	if lm.stopped != nil {
+		lm.stopped <- true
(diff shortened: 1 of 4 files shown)


jiayliu:
@milkypostman

src/collider/collider/logmanager.go
```diff
@@ -0,0 +1,143 @@
...
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"path"
+	"strings"
+	"time"
+)
+
+// LogManager redirects log output to files, one for each |cycle|, and deletes files that are last modified before the retention period.
```
tkchin: With the disclaimer that I don't know Go style - this comment seems to exceed 80 chars.

src/collider/collider/logmanager.go
```diff
@@ -0,0 +1,143 @@
...
+func NewLogManager(d string, c time.Duration, r time.Duration) *LogManager {
+	if d == "" {
+		log.Printf("Invalid log dir")
+		return nil
+	}
+	if c >= r {
+		log.Printf("Log retention time must be greater than log cycle")
+		return nil
+	}
+	return &LogManager{
```
tkchin: Can you run gofmt if you haven't already? Probably the best way to deal with style in the absence of someone with readability. Thanks.

milkypostman:
I don't think we should build this functionality into the server. We can accomplish the same thing with the following two server changes.

1. run collider with `collidermain >> /path/to/log/file`
2. Add a `collider` file to `/etc/logrotate.d` that contains the following,

```
/path/to/log/file {
    daily
    compress
    copytruncate
    dateext
    missingok
    notifempty
    rotate 10
    sharedscripts
}
```

jiayliu:
I haven't figured out how to rotate the log without restarting Collider. If
logrotate removes the old log and Collider is not restarted, no more
logging will be saved.

On Fri, Feb 20, 2015 at 11:24 AM, Donald Ephraim Curtis <

> I don't think we should build this functionality into the server. We can
> accomplish the same thing with the following two server changes.
>
>    1. run collider with collidermain >> /path/to/log/file
>    2. Add a collider file to /etc/logrotate.d that contains the following,
>
> /path/to/log/file {
>     daily
>     compress
>     copytruncate
>     dateext
>     missingok
>     notifempty
>     rotate 10
>     sharedscripts
> }
>
> —
> <https://github.com/webrtc/apprtc/pull/85#issuecomment-75302617>.
>



-- 
Jiayang Liu

milkypostman:
This shouldn't be an issue because we are using the `copytruncate` and using a `>>` redirect from the shell. I am happy to write an application that runs a loop and logs things but from my [limit research](http://unix.stackexchange.com/questions/42544/does-redirecting-output-to-a-file-apply-a-lock-on-the-file) (i see this same advice elsewhere) it won't be an issue. 

In other words, bash/sh will be directing output from `collidermain` to `/path/to/log/file` and when logrotate runs it will copy `file` and truncate it.

The other option is to add a command to logrotate to restart collider in the case of log rotation but I don't think that's necessary and it would wreck some call experience right? Like if people are in a call when collider restarts?

jiayliu:
SG

On Fri, Feb 20, 2015 at 12:02 PM, Donald Ephraim Curtis <

> This shouldn't be an issue because we are using the copytruncate and
> using a >> redirect from the shell. I am happy to write an application
> that runs a loop and logs things but from my limit research
> <http://unix.stackexchange.com/questions/42544/does-redirecting-output-to-a-file-apply-a-lock-on-the-file>
> (i see this same advice elsewhere) it won't be an issue.
>
> In other words, bash/sh will be directing output from collidermain to
> /path/to/log/file and when logrotate runs it will copy file and truncate
> it.
>
> The other option is to add a command to logrotate to restart collider in
> the case of log rotation but I don't think that's necessary and it would
> wreck some call experience right? Like if people are in a call when
> collider restarts?
>
> —
> <https://github.com/webrtc/apprtc/pull/85#issuecomment-75309196>.
>



-- 
Jiayang Liu

jiayliu:
Let's go with the logrotate solution

PR #88, opened 2015-02-20
Fix apiauth to insert third_party modules first.

milkypostman:
The new appengine sdk release added oauth2client of its own which was conflicting with ours.

### DIFF
--- a/src/app_engine/apiauth.py
+++ b/src/app_engine/apiauth.py
@@ -6,9 +6,10 @@ import json
 import os
 import sys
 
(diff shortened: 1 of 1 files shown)


src/app_engine/apiauth.py
```diff
@@ -6,9 +6,9 @@
 import os
 import sys
 
-sys.path.append(os.path.join(os.path.dirname(__file__), 'third_party'))
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), 'third_party'))
```
jiayliu: Please comment why we do this

jiayliu:
lgtm %nit

PR #89, opened 2015-02-21
Updated the third-party commit flow and updated the httplib2 code to new version based on this flow.

milkypostman:
After some discussion about the breakage earlier, caused by conflicts between our third-party and the appengine third-party, I decided to try out their suggestions. It's based on this post http://blog.jonparrott.com/managing-vendored-packages-on-app-engine/ about managing "vendored" packages.

It seems like the official way to install the `google-api-python-client` is through `pip` and the `pypi` repository. So I want to do that now. Eventually we'll update to `google-api-python-client==1.3.1` which is the newest but for now leaving at `1.2` and just using the new flow.

Mainly this flow means that we pull in `httplib2==0.9` instead of `0.8`.


chuckhays:
That looks like a lot of magic, but as far as I can tell, LGTM.

milkypostman:
haha, what's magic?

chuckhays:
Just the egg files, no issues.

milkypostman:
yeah. if you look at the second item in the readme there is a bug when updating packages. normally you don't have to delete the old packages and pip gets the information from the egg files. but since there is a bug the egg files are actually going to get wiped each time we update since we delete all the files anyways. 

I figured I'd just do it the right way anyways. egg files are a weird Python thing I've never really understood.

PR #91, opened 2015-02-23
Merge changes from master

tkchin:
@jiayliu @milkypostman PTAL

Unit tests pass and open room connects. Most of the conflicts were analytics related, let me know if I missed anything.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,10 @@
-node_modules
-.DS_Store
 *.pyc
-validation-report.json
-validation-status.json
+.DS_Store
 bower_components
 google_appengine*
-webtest-master*
+node_modules
 out
+secrets.json
+validation-report.json
+validation-status.json
+webtest-master*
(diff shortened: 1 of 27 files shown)


tkchin:
FYI 871c1a3b9fb46ab9daa03aa1ff4de50fe82e4abf has the list of files with conflicts.

milkypostman:
will take a look this afternoon in more detail, but it looks like it makes sense :)

tkchin:
Ok lmk. Can't run UTs anymore in this branch due to to API auth thing we ran into until this merge goes in, which blocks any other CLs on this branch.

milkypostman:
The only weird thing I see is that we now have `test_uti.py` and `test_utilities.py`. I'm fine with this as it will make future commits easier to merge.

It's a bit hard to decrypt this but I checked that the things I added directly to `apprtc.py` were in the appropriate area after merge and all other files were the same.

LGTM

tkchin:
Thanks for looking! Will merge if direct calling scenario works.

PR #92, opened 2015-02-24
Use prober results to return working WS server.

chuckhays:
This implements a simple strategy of returning whichever collider instance has been up the longest.

For example, assume we have servers A and B which are both up, and GAE is returning A
If A is reported as down by prober, GAE will start returning B. GAE will continue serving B even if A comes back up. B will returned until B if reported as down by problem.

This will help avoid the situation where clients within a room are in different collider instances. It does not get us any load balancing benefits. 

If we want to do something more advanced here to ensure clients get into the right collider instance together, the next steps would be:
1. When the first client joins a room, store the collider instance assigned to the client along with the room data.
2. When the second client joins a room, ensure we return that same collider instance.
3. Client would need to be updated to reconnect to the socket if the wss address returned by /join is different than the one it started with.

The situation this would prevent would be:
GAE is currently returning A,
Client 2 loads /, gets A in the initial params
A goes down, B is now returned
Client 1 loads 1/, gets B, joins room, gets assigned B, connects to B
Client 2 now joins a room, but connects to A because we connect before /join. When join returns B, client 2 needs to disconnect from A (possibly failed anyway) and connect to B.

I suspect the effort for the more advanced tracking/reconnecting is not worth it at this time.

@jiayliu @tkchin

### DIFF
--- a/src/app_engine/apprtc.py
+++ b/src/app_engine/apprtc.py
@@ -115,18 +115,28 @@ def append_url_arguments(request, link):
   return link
 
 def get_wss_parameters(request):
-  ws_host_port_pair = request.get('wshpp')
-  ws_tls = request.get('wstls')
+  wss_host_port_pair = request.get('wshpp')
+  wss_tls = request.get('wstls')
 
-  if not ws_host_port_pair:
-    ws_host_port_pair = constants.WSS_HOST_PORT_PAIR
+  if not wss_host_port_pair:
+    # Attempt to get a wss server from the status provided by prober,
+    # if that fails, use fallback value.
+    memcache_client = memcache.Client()
+    wss_active_host = memcache_client.get(constants.WSS_HOST_ACTIVE_HOST_KEY)
+    if wss_active_host in constants.WSS_HOST_PORT_PAIRS:
+      wss_host_port_pair = wss_active_host
+    else:
+      logging.warning(
+          'Invalid or no value returned from memcache, using fallback: '
+          + json.dumps(wss_active_host))
+      wss_host_port_pair = constants.WSS_HOST_PORT_PAIRS[0]
 
-  if ws_tls and ws_tls == 'false':
-    wss_url = 'ws://' + ws_host_port_pair + '/ws'
-    wss_post_url = 'http://' + ws_host_port_pair
+  if wss_tls and wss_tls == 'false':
+    wss_url = 'ws://' + wss_host_port_pair + '/ws'
+    wss_post_url = 'http://' + wss_host_port_pair
   else:
-    wss_url = 'wss://' + ws_host_port_pair + '/ws'
-    wss_post_url = 'https://' + ws_host_port_pair
+    wss_url = 'wss://' + wss_host_port_pair + '/ws'
+    wss_post_url = 'https://' + wss_host_port_pair
   return (wss_url, wss_post_url)
 
 def get_version_info():
--- a/src/app_engine/apprtc_test.py
+++ b/src/app_engine/apprtc_test.py
@@ -5,12 +5,21 @@ import time
 import unittest
 
 import webtest
-from google.appengine.ext import testbed
 
 import analytics
 import apprtc
+import constants
+import probers
 from test_util import CapturingFunction
 from test_util import ReplaceFunction
+
+from google.appengine.api import memcache
+from google.appengine.ext import testbed
+
+
+class MockRequest(object):
+  def get(self, key):
+    return None
 
 
 class AppRtcUnitTest(unittest.TestCase):
@@ -131,6 +140,49 @@ class AppRtcPageHandlerTest(unittest.TestCase):
 
     self.makePostRequest('/leave/' + room_id + '/' + caller_id)
     self.makePostRequest('/leave/' + room_id + '/' + callee_id)
+
+  def setWssHostStatus(self, index1, status1, index2, status2):
+    probing_results = {}
+    probing_results[constants.WSS_HOST_PORT_PAIRS[index1]] = {
+        constants.WSS_HOST_IS_UP_KEY: status1
+    }
+    probing_results[constants.WSS_HOST_PORT_PAIRS[index2]] = {
+        constants.WSS_HOST_IS_UP_KEY: status2
+    }
+    probers.ProbeColliderPage().store_instance_state(probing_results)
+
+  def verifyRequest(self, expectedIndex):
+    request = MockRequest()
+    wss_url, wss_post_url = apprtc.get_wss_parameters(request)
+    self.assertIn(constants.WSS_HOST_PORT_PAIRS[expectedIndex], wss_url)
+    self.assertIn(constants.WSS_HOST_PORT_PAIRS[expectedIndex], wss_post_url)
+
+  def testGetWssHostParameters(self):
+    request = MockRequest()
+    # With no status set, should use fallback.
+    self.verifyRequest(0)
+
+    # With an invalid value in memcache, should use fallback.
+    memcache_client = memcache.Client()
+    memcache_client.set(constants.WSS_HOST_ACTIVE_HOST_KEY, 'abc')
+    self.verifyRequest(0)
+
+    # With an invalid value in memcache, should use fallback.
+    memcache_client = memcache.Client()
+    memcache_client.set(constants.WSS_HOST_ACTIVE_HOST_KEY, ['abc', 'def'])
+    self.verifyRequest(0)
+
+    # With both hosts failing, should use fallback.
+    self.setWssHostStatus(0, False, 1, False)
+    self.verifyRequest(0)
+
+    # Second host passing.
+    self.setWssHostStatus(0, False, 1, True)
+    self.verifyRequest(1)
+
+    # Both hosts passing, but second host for longer.
+    self.setWssHostStatus(1, True, 0, True)
+    self.verifyRequest(1)
 
 
 if __name__ == '__main__':
--- a/src/app_engine/constants.py
+++ b/src/app_engine/constants.py
@@ -8,6 +8,7 @@ import json
 import os
 
 ROOM_MEMCACHE_EXPIRATION_SEC = 60 * 60 * 24
+MEMCACHE_RETRY_LIMIT = 100
 
 LOOPBACK_CLIENT_ID = 'LOOPBACK_CLIENT_ID'
 
@@ -15,7 +16,13 @@ TURN_BASE_URL = 'https://computeengineondemand.appspot.com'
 TURN_URL_TEMPLATE = '%s/turn?username=%s&key=%s'
 CEOD_KEY = '4080218913'
 
-WSS_HOST_PORT_PAIR = 'apprtc-ws.webrtc.org:443'
+WSS_HOST_PORT_PAIRS = ['apprtc-ws.webrtc.org:443', 'apprtc-ws-2.webrtc.org:443']
+WSS_HOST_ACTIVE_HOST_KEY = 'wss_host_active_host'
+
+WSS_HOST_INSTANCE_KEY = 'instance'
+WSS_HOST_IS_UP_KEY = 'is_up'
+WSS_HOST_STATUS_CODE_KEY = 'status_code'
+WSS_HOST_ERROR_MESSAGE_KEY = 'error_message'
 
 RESPONSE_ERROR = 'ERROR'
 RESPONSE_ROOM_FULL = 'FULL'
--- a/src/app_engine/probers.py
+++ b/src/app_engine/probers.py
@@ -12,7 +12,9 @@ import logging
 import json
 import numbers
 import webapp2
+
 from google.appengine.api import mail
+from google.appengine.api import memcache
 from google.appengine.api import urlfetch
 
 def send_alert_email(tag, message):
@@ -94,18 +96,65 @@ class ProbeCEODPage(webapp2.RequestHandler):
     self.handle_ceod_response(error_message, status_code)
 
 class ProbeColliderPage(webapp2.RequestHandler):
-  def handle_collider_response(self, error_message, status_code):
-    self.response.set_status(status_code)
+  def handle_collider_response(
+      self, error_message, status_code, collider_instance):
+    result = {
+        constants.WSS_HOST_STATUS_CODE_KEY: status_code
+    }
     if error_message is not None:
-      send_alert_email('Collider Error', error_message)
+      send_alert_email('Collider Error (' + collider_instance + ')',
+                       error_message)
 
       logging.warning('Collider prober error: ' + error_message)
-      self.response.out.write(error_message)
+      result[constants.WSS_HOST_ERROR_MESSAGE_KEY] = error_message
+      result[constants.WSS_HOST_IS_UP_KEY] = False
     else:
-      self.response.out.write('Success!')
+      result[constants.WSS_HOST_IS_UP_KEY] = True
+    return result
+
+  def store_instance_state(self, probing_results):
+    # Store an active collider host to memcache to be served to clients.
+    # If the currently active host is still up, keep it. If not, pick a 
+    # new active host that is up.
+    memcache_client = memcache.Client()
+    for retries in xrange(constants.MEMCACHE_RETRY_LIMIT):
+      active_host = memcache_client.gets(constants.WSS_HOST_ACTIVE_HOST_KEY)
+      if active_host is None:
+        memcache_client.set(constants.WSS_HOST_ACTIVE_HOST_KEY, '')
+        active_host = memcache_client.gets(constants.WSS_HOST_ACTIVE_HOST_KEY)
+      active_host = self.create_collider_active_host(active_host,
+                                                     probing_results)
+      if memcache_client.cas(constants.WSS_HOST_ACTIVE_HOST_KEY, active_host):
+        logging.info('collider active host saved to memcache: ' +
+                     str(active_host))
+        break
+      logging.warning('retry # ' + str(retries) + ' to set collider status')
+
+  def create_collider_active_host(self, old_active_host, probing_results):
+    # If the old_active_host is still up, keep it. If not, pick a new active
+    # host that is up.
+    try:
+      if (old_active_host in probing_results and
+          probing_results[old_active_host].get(
+              constants.WSS_HOST_IS_UP_KEY, False)):
+        return old_active_host
+    except TypeError:
+      pass
+    for instance in probing_results:
+      if probing_results[instance].get(constants.WSS_HOST_IS_UP_KEY, False):
+        return instance
+    return None
 
   def get(self):
-    url = 'https://' + constants.WSS_HOST_PORT_PAIR + '/status';
+    results = {}
+    for collider_instance in constants.WSS_HOST_PORT_PAIRS:
+      results[collider_instance] = self.probe_collider_instance(
+          collider_instance)
+    self.response.write(json.dumps(results, indent=2, sort_keys=True))
+    self.store_instance_state(results)
+
+  def probe_collider_instance(self, collider_instance):
+    url = 'https://' + collider_instance + '/status';
 
     error_message = None
     result = None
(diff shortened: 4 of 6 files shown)


src/app_engine/probers.py
```diff
@@ -12,13 +12,18 @@
...
+from google.appengine.api import memcache
 from google.appengine.api import urlfetch
 
-def send_alert_email(tag, message):
+def send_alert_email(tag, message, instance=None):
   receiver = 'tachyon-eng@google.com'
   sender_address = 'AppRTC Notification <jiayl@google.com>'
-  subject = 'AppRTC Prober Alert: ' + tag
+  subject = 'AppRTC Prober Alert'
+  if (instance != None):
```
tkchin: Run linter please.

src/app_engine/probers.py
```diff
@@ -94,18 +99,74 @@ def get(self):
...
+          collider_status, results)
+      if memcache_client.cas(constants.WSS_HOST_STATUS_LIST_KEY,
+                             collider_status):
+        logging.info('collider status saved to memcache')
+        break
+      logging.warning('retry # ' + str(retries) + ' to set collider status')
+
+  def create_collider_status_list(self, existing_record, results):
+    # Keep only the instance name and status.
+    sorted_results = [{
```
tkchin: If you store tuples instead you can avoid all this sorting logic. You can use WSS_HOST_INSTANCE_INDEX, WSS_HOST_STATUS_INDEX instead.

src/app_engine/apprtc.py
```diff
@@ -119,7 +119,22 @@ def get_wss_parameters(request):
   ws_tls = request.get('wstls')
 
   if not ws_host_port_pair:
-    ws_host_port_pair = constants.WSS_HOST_PORT_PAIR
+    memcache_client = memcache.Client()
```
tkchin: Why memcache and not ndb?

src/app_engine/constants.py
```diff
@@ -15,7 +16,13 @@
 TURN_URL_TEMPLATE = '%s/turn?username=%s&key=%s'
 CEOD_KEY = '4080218913'
 
-WSS_HOST_PORT_PAIR = 'apprtc-ws.webrtc.org:443'
+WSS_HOST_PORT_PAIRS = ['apprtc-ws.webrtc.org:443', 'apprtc-ws-2.webrtc.org:443']
+WSS_HOST_STATUS_LIST_KEY = 'wss_host_status_list'
+
+WSS_HOST_INSTANCE_KEY = 'instance'
+WSS_HOST_STATUS_KEY = 'status'
```
jiayliu: I find 'status' and 'status_code' are confusing. Maybe 'status' should be 'is_up' or 'is_healthy'

src/app_engine/constants.py
```diff
@@ -15,7 +16,13 @@
...
+WSS_HOST_STATUS_CODE_KEY = 'status_code'
```
jiayliu: How is 'status_code' and 'error_message' used? Do we need to store them?

src/app_engine/probers.py
```diff
@@ -12,13 +12,18 @@
...
 from google.appengine.api import urlfetch
 
-def send_alert_email(tag, message):
+def send_alert_email(tag, message, instance=None):
   receiver = 'tachyon-eng@google.com'
   sender_address = 'AppRTC Notification <jiayl@google.com>'
-  subject = 'AppRTC Prober Alert: ' + tag
+  subject = 'AppRTC Prober Alert'
+  if (instance != None):
+    subject += ' (' + instance + ') '
```
jiayliu: Why not just put 'instance' in the tag?

src/app_engine/probers.py
```diff
@@ -94,18 +99,74 @@ def get(self):
...
+    # Sort by order in existing record and status.
+    sorted_results.sort(
+        key=lambda server: self.get_index(
+            existing_record,
+            server[constants.WSS_HOST_INSTANCE_KEY]))
+    sorted_results.sort(
+        key=lambda server: server[constants.WSS_HOST_STATUS_KEY], reverse=True)
+    return sorted_results
+
+  def get_index(self, input_list, instance):
```
jiayliu: nit: comment what this method does

src/app_engine/probers.py
```diff
@@ -94,18 +99,74 @@ def get(self):
...
 
   def get(self):
-    url = 'https://' + constants.WSS_HOST_PORT_PAIR + '/status';
+    results = []
+    for collider_instance in constants.WSS_HOST_PORT_PAIRS:
+      results.append(self.test_collider_instance(collider_instance))
+    self.response.write(json.dumps(results, indent=2, sort_keys=True))
+    self.store_instance_state(results)
+
+  def test_collider_instance(self, collider_instance):
```
jiayliu: nit: probe_collider_instance?

src/app_engine/probers_test.py
```diff
@@ -0,0 +1,142 @@
...
+  def testGetIndex(self):
+    server_list = [
+        self.createEntry('server2', True),
+        self.createEntry('server3', True),
+        self.createEntry('serverD', True),
+        self.createEntry('serverB', True),
+        self.createEntry('serverC', False, 500, FAKE_ERROR_MESSAGE),
+        self.createEntry('server1', False, 500, FAKE_ERROR_MESSAGE)
+    ]
+    self.assertEqual(
```
jiayliu: Can we put the asserts in a loop to be more concise?

src/app_engine/probers.py
jiayliu: Or in the 'message'. I'd prefer to avoid extra argument that is not common to CEOD and collider.

src/app_engine/apprtc.py
jiayliu: I personally think memcache is easier to use than ndb, and faster.

src/app_engine/probers.py
chuckhays: I linted all the new code. I was concerned about merging to the other branch given that we've linted a bunch of the files there. This change will already require some hand changes when merging.

src/app_engine/constants.py
chuckhays: Good idea.

src/app_engine/constants.py
chuckhays: Those are part of what is returned via http when you call /probe/collider. More useful as a status/debugging tool. We could store them, but nothing would be consuming them currently.

src/app_engine/probers.py
chuckhays: Formatting was the only reason, I'll change it.

src/app_engine/constants.py
jiayliu: It seems nothing is consuming it even in memory. Can we just remove them?

chuckhays: Prior to my change, the http request would just return the error_message if there was one, and set the http return status. Now it returns a json doc with the instance name, error message, http status code and up/down status. If we don't care about returning the error message via the http call, then yes, we can.

jiayliu: OK, handle_collider_response removed set_status and removed writing the error message to response, that's why these values are not used anywhere anymore. I think we should reserve the status code and response message.

chuckhays: They are still written out on line 167. handle_collider_response creates a result object for each instance, which are added to an array. The array is written out on 167.

Do you think we should do something different? I was trying to keep a human readable output so that you can call /probe/collider from a browser and see the current status.

src/app_engine/apprtc.py
tkchin: Ok. Reason I asked was more because with ndb could've added filter / order in query so we don't have to do sorting stuff below.

src/app_engine/constants.py
jiayliu: OK, I see it now. Issue solved.

On Wed, Feb 25, 2015 at 2:28 PM, Chuck Hays <notifications@github.com>
wrote:

> In src/app_engine/constants.py
> <https://github.com/webrtc/apprtc/pull/92#discussion_r25386200>:
>
> > @@ -15,7 +16,13 @@
> >  TURN_URL_TEMPLATE = '%s/turn?username=%s&key=%s'
> >  CEOD_KEY = '4080218913'
> >
> > -WSS_HOST_PORT_PAIR = 'apprtc-ws.webrtc.org:443'
> > +WSS_HOST_PORT_PAIRS = ['apprtc-ws.webrtc.org:443', 'apprtc-ws-2.webrtc.org:443']
> > +WSS_HOST_STATUS_LIST_KEY = 'wss_host_status_list'
> > +
> > +WSS_HOST_INSTANCE_KEY = 'instance'
> > +WSS_HOST_STATUS_KEY = 'status'
> > +WSS_HOST_STATUS_CODE_KEY = 'status_code'
>
> They are still written out on line 167. handle_collider_response creates a
> result object for each instance, which are added to an array. The array is
> written out on 167.
>
> Do you think we should do something different? I was trying to keep a
> human readable output so that you can call /probe/collider from a browser
> and see the current status.
>
> —
> <https://github.com/webrtc/apprtc/pull/92/files#r25386200>.
>



-- 
Jiayang Liu

src/app_engine/apprtc.py
```diff
@@ -119,7 +119,22 @@ def get_wss_parameters(request):
...
+    wss_host_status_list = memcache_client.get(
+        constants.WSS_HOST_STATUS_LIST_KEY)
+    # Return the first server in the list that is up, which is the wss host
+    # that has been up the longest.
+    if wss_host_status_list is not None:
+      wss_host_status_list = [host for host in wss_host_status_list
```
milkypostman: Another way to do this which I think is a bit more Pythonic is,

```python
try:
  ws_host = (host for host in wss_host_status_list if host[constants.WSS_HOST_IS_UP_KEY]).next()
  ws_host_port_pair = ws_host[constants.WSS_HOST_INSTANCE_KEY]
except StopIteration:
  logging.warning('no wss servers are reported as up, using fallback')
  ws_host_port_pair = constants.WSS_HOST_PORT_PAIRS[0]
```

My arguments:

1. using a try is better in cases where you assume it will succeed and I think in *general* we assume this will succeed; i.e., one host should be up.
2. It avoids iteration through the entire list of hosts. Doesn't matter with our small list I guess.

chuckhays: I'm still learning the Pythonic way, so if this fits, I'm all for it. I miss .FirstOrDefault() clauses.

chuckhays: Take a look at the latest commit and tell me what you think, I modified your suggestion slightly to use next() with a default to avoid the try/catch.

chuckhays:
@milkypostman Switched to what we discussed, also added a couple test cases to cover the invalid data scenario.

src/app_engine/probers.py
```diff
@@ -94,18 +96,82 @@ def get(self):
...
+      if memcache_client.cas(constants.WSS_HOST_STATUS_LIST_KEY,
+                             collider_status):
+        logging.info('collider status saved to memcache')
+        break
+      logging.warning('retry # ' + str(retries) + ' to set collider status')
+
+  def create_collider_status_list(self, existing_record, results):
+    # Keep only the instance name and status.
+    sorted_results = [{
+        constants.WSS_HOST_INSTANCE_KEY:
```
milkypostman:
```python
existing = dict((instance[constants.WSS_HOST_INSTANCE_KEY], idx) for idx, instance in ienum(existing_record))
numbered_results = [(existing.get(obj[constants.WSS_HOST_INSTANCE_KEY], len(results)), obj) for obj in results]
return [obj for _, obj in sorted(numbered_results)]
```
something awesome but people can ignore if they are not @chuckhays

chuckhays:
@milkypostman @tkchin 
I simplified what was being stored. Now memcache only stores the active host, since that is all we were using anyway. The update step now can just check if the active host is still active, or if not, pick one that is up and store that.

src/app_engine/apprtc.py
```diff
@@ -115,18 +115,28 @@ def append_url_arguments(request, link):
   return link
 
 def get_wss_parameters(request):
-  ws_host_port_pair = request.get('wshpp')
+  wss_host_port_pair = request.get('wshpp')
   ws_tls = request.get('wstls')
```
tkchin: nit: wss_tls too

src/app_engine/apprtc_test.py
```diff
@@ -6,12 +6,19 @@
 
 import webtest
 from google.appengine.ext import testbed
+from google.appengine.api import memcache
```
tkchin: nit: import order / relint

src/app_engine/probers.py
```diff
@@ -12,7 +12,9 @@
 import json
 import numbers
 import webapp2
+import sys
```
tkchin: nit: relint

src/app_engine/probers.py
```diff
@@ -94,18 +96,64 @@ def get(self):
...
+
+  def store_instance_state(self, probing_results):
+    # Store a list with status for each collider host.
+    # The list should be in order of decreasing uptime (the first server in
+    # the list has been up the longest).
+    memcache_client = memcache.Client()
+    for retries in xrange(constants.MEMCACHE_RETRY_LIMIT):
+      active_host = memcache_client.gets(constants.WSS_HOST_ACTIVE_HOST_KEY)
+      if active_host is None:
+        memcache_client.set(constants.WSS_HOST_ACTIVE_HOST_KEY, '')
```
tkchin: maybe set default to {} instead to avoid TypeError catch below

src/app_engine/probers.py
```diff
@@ -94,18 +96,64 @@ def get(self):
...
+      if memcache_client.cas(constants.WSS_HOST_ACTIVE_HOST_KEY, active_host):
+        logging.info('collider status saved to memcache')
+        break
+      logging.warning('retry # ' + str(retries) + ' to set collider status')
+
+  def create_collider_active_host(self, old_active_host, probing_results):
+    # If the old_active_host is still up, keep it. If not, pick a new active
+    # host that is up.
+    try:
+      if (old_active_host in probing_results and
```
tkchin: if using empty dictionary you can do
still_alive = old_active_host in probing_results and probing_result[old_active_host].get()

src/app_engine/probers.py
```diff
@@ -94,18 +96,64 @@ def get(self):
...
-      self.response.out.write(error_message)
+      result[constants.WSS_HOST_ERROR_MESSAGE_KEY] = error_message
+      result[constants.WSS_HOST_IS_UP_KEY] = False
     else:
-      self.response.out.write('Success!')
+      result[constants.WSS_HOST_IS_UP_KEY] = True
+    return result
+
+  def store_instance_state(self, probing_results):
+    # Store a list with status for each collider host.
```
tkchin: Update comment?

src/app_engine/probers.py
```diff
@@ -94,18 +96,64 @@ def get(self):
...
+    except StopIteration:
+      return None
 
   def get(self):
-    url = 'https://' + constants.WSS_HOST_PORT_PAIR + '/status';
+    results = {}
+    for collider_instance in constants.WSS_HOST_PORT_PAIRS:
+      results[collider_instance] = self.probe_collider_instance(
+          collider_instance)
+    self.response.write(json.dumps(results, indent=2, sort_keys=True))
```
tkchin: Do you also want to write active host?

src/app_engine/probers.py
tkchin: nevermind :p

src/app_engine/probers.py
tkchin: ignore dictionary comment. However, can still use the get
probing_result[old_active_host].get(constants.WSS_HOST_IS_UP_KEY, False)

src/app_engine/probers.py
```diff
@@ -94,18 +96,64 @@ def get(self):
...
+    # If the old_active_host is still up, keep it. If not, pick a new active
+    # host that is up.
+    try:
+      if (old_active_host in probing_results and
+          probing_results[old_active_host][constants.WSS_HOST_IS_UP_KEY]):
+        return old_active_host
+    except TypeError:
+      pass
+    try:
+      return (instance for instance in probing_results
```
tkchin: why do we need the try/catch + next? Maybe better to just write an explicit loop:
for hostname, probing_result in probing_results.items():
  if probing_result.get(constants.WSS_HOST_IS_UP_KEY):
    return hostname

tkchin:
Thanks for the simplification :)

src/app_engine/probers.py
chuckhays: @milkypostman convinced me that was the way to go in python, arguments against? I didn't like using the try/except for flow control, but apparently that is "the way"

src/app_engine/probers.py
chuckhays: So my concern was old_active_host could be something like a list from bad data, which is caught by the TypeError, am I missing what is gained by using .get here?

src/app_engine/probers.py
chuckhays: Probably a good idea.

src/app_engine/probers.py
chuckhays: Good catch.

src/app_engine/probers.py
tkchin: The idiomatic way of looping over a sequence in python is a for loop. If this were the way to go in Python we'd we writing try/catch + next for looping over a list.

next() is more for generator functions / expressions that you don't know the length of beforehand. In that scenario StopIteration makes sense given Python's "assume it works" style.

Also I meant iteritems on the dict. And yes, need a return None at the end.

src/app_engine/probers.py
tkchin: If it's bad data it means we wrote code wrong somehow and should trigger an assert / throw an exception. That is to say, we should try/catch expected exceptions, but we should let unexpected exceptions trigger. This should be storing the result from handle_collider_response, which we know is a dictionary for sure (or at least it should be). Using get protects us if for some reason the key doesn't exist (through it should since we always set it above).

chuckhays: The exception in this case would be from the left side of the if. It is admittedly far fetched that memcache would return corrupted data, or something unexpected other than None, but that was what I was thinking of.

src/app_engine/probers.py
milkypostman: @chuckhays i have created a monster! :smiling_imp:  I forgot that I saw this but `next` provides a default value
```python
return next((instance for instance in probing_results 
          if probing_results[instance][constants.WSS_HOST_IS_UP_KEY]), None)
```
and you could do,
```python
up_instances = (instance for instance in probing_results 
                     if probing_results[instance][constants.WSS_HOST_IS_UP_KEY])
return next(up_instances, None)
```
but @tkchin may have a point that,
```python
for instance in probing_results:
  if probing_results[instance][constants.WSS_HOST_IS_UP_KEY]:
    return instance
return None
```
is more readable. In the previous situation we were discussing where we didn't know the state of our variables and you were catching `None` and an empty list, I would make the argument it was cleaner.

tkchin:
lgtm

src/app_engine/probers_test.py
```diff
@@ -0,0 +1,97 @@
...
+  def verifyActiveHost(self, old_active_host, probing_results,
+                       possible_active_hosts):
+    new_active_host = probers.ProbeColliderPage().create_collider_active_host(
+        old_active_host,
+        probing_results)
+    self.assertIn(new_active_host, possible_active_hosts)
+
+  def createEntry(
+      self,
+      status,
```
jiayliu: nit: is_up

src/app_engine/probers_test.py
```diff
@@ -0,0 +1,97 @@
...
+                          possible_active_hosts)
+
+  def verifyActiveHost(self, old_active_host, probing_results,
+                       possible_active_hosts):
+    new_active_host = probers.ProbeColliderPage().create_collider_active_host(
+        old_active_host,
+        probing_results)
+    self.assertIn(new_active_host, possible_active_hosts)
+
+  def createEntry(
```
jiayliu: nit: move the test* methods together

jiayliu:
Could you also update cron.yaml to run the probers every 5min?

lgtm %nits

PR #99, opened 2015-03-03
figure out the local type preference of the active candidate

fippo:
basically https://github.com/webrtc/samples/pull/281 -- only for local candidates. Remote candidate guessing yielded some weird results, need to look again.

### DIFF
--- a/src/web_app/js/infobox.js
+++ b/src/web_app/js/infobox.js
@@ -8,8 +8,8 @@
 
 /* More information about these options at jshint.com/docs/options */
 
-/* globals computeBitrate, computeE2EDelay, extractStatAsInt, getStatsReport,
-   iceCandidateType, computeRate */
+/* globals computeBitrate, computeE2EDelay, extractStatAsInt,
+   formatTypePreference, getStatsReport, iceCandidateType, computeRate */
 /* exported InfoBox */
 
 'use strict';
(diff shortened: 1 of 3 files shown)


src/web_app/js/sdputils.js
```diff
@@ -34,6 +35,20 @@ function iceCandidateType(candidateStr) {
   return candidateStr.split(' ')[7];
 }
 
+// Turns the local type preference into a human-readable string.
+// Note that this mapping is browser-specific.
+function formatTypePreference(pref) {
```
jiayliu: Should this go to adapter.js and make it work for FF as well?

fippo: good question... I don't think it is generally useful enough, but it should have a switch for Firefox support here

fippo: oh... it's currently not called in Firefox at all since that doesn't support the stats code (yet). Anyway,   it might in the future.

KaptenJansson:
fippo@ anything else needed here?

src/web_app/js/infobox.js
```diff
@@ -115,8 +115,19 @@ InfoBox.prototype.updateInfoDiv = function() {
...
+      if (localCandId) {
+        localCand = getStatsReport(this.stats_, 'localcandidate', 'id',
+            localCandId);
+        if (localCand) {
+          localTypePref = localCand.stat('priority') >> 24;
+        }
+      }
       contents += this.buildLine_('LocalAddr', localAddr +
-          ' (' + localAddrType + ')');
+          ' (' + localAddrType + (localTypePref !== undefined ?
```
KaptenJansson: Should this be `(typeof localTypePref !== 'undefined'` ?

fippo: i don't think this makes a difference. Actually this might just do a non-false check, i.e. 
+ (localTypePref ? formatTypePref(localTypePref) : '')

fippo:
@KaptenJansson firefox support -- but that requires spec-style getStats first :-)

KaptenJansson:
> @KaptenJansson firefox support -- but that requires spec-style getStats first :-)

As in spec style getStats in Apprtc or Firefox (thought they were close to spec?)?

Or are you referring to the same as you mentioned in https://github.com/webrtc/apprtc/pull/162#issuecomment-122115953?

KaptenJansson:
@fippo ping

fippo:
err... requires removing all the .stat calls. See https://github.com/webrtc/samples/pull/606 for an example.
I'm not going to do that however :-)

src/web_app/js/infobox.js
KaptenJansson: Since other files in the apprtc repo have it the way you are doing it, it's fine.

KaptenJansson:
LGTM

fippo:
turnudp
-- thanks!

PR #100, opened 2015-03-03
Publish to chrome app store

chuckhays:
Adds required images so we can publish to the store and get a stable extension id during development. Please excuse the developer art.
Add manifest key so extension id is stable during development.

@jiayliu


jiayliu:
Thx! lgtm

PR #104, opened 2015-03-04
Analytics page should report the host.

milkypostman:
Analytics changed to using named arguments in hopes of avoiding any confusions.


chuckhays:
LGTM. 

HOST is defined in the BasePageHandlerTest in the refactored test code in apprc-t, should we be doing this work in the branch instead of main to take advantage of the refactoring we did there?

jiayliu:
lgtm

milkypostman:
@chuckhays probably we need to backport that change but I will do that when I get a chance in another CL.

PR #106, opened 2015-03-04
Increase prober urlfetch deadline to 30s.

tkchin:


tkchin:
@jiayliu @chuckhays PTAL

jiayliu:
OOC, what's the default timeout?

jiayliu:
nv my last comment. lgtm

PR #108, opened 2015-03-05
Restart collider instance when prober fails

jiayliu:
@milkypostman @chuckhays

### DIFF
--- a/src/app_engine/compute_page.py
+++ b/src/app_engine/compute_page.py
@@ -68,7 +68,7 @@ class ComputePage(webapp2.RequestHandler):
     """Implementation for restart action.
 
     Args:
-      instance: Name of the instance to start.
+      instance: Name of the instance to restart.
       zone: Name of the zone the instance belongs to.
 
     """
@@ -114,9 +114,11 @@ class ComputePage(webapp2.RequestHandler):
           project=app_identity.get_application_id(),
           instance=instance,
           zone=zone).execute()
-    elif status != COMPUTE_STATUS_RUNNING:
-      # Instance is in an intermediate state: PROVISIONING, STAGING,
-      # STOPPING. Thus, requeue the task.
+
+    if status != COMPUTE_STATUS_RUNNING:
+      # If in an intermediate state: PROVISIONING, STAGING, STOPPING, requeue
+      # the task to check back later. If in TERMINATED state, also requeue the
+      # task since the start attempt may fail and we should retry.
       enqueue_start_task(instance, zone)
 
   def _compute_status(self, instance, zone):
--- a/src/app_engine/compute_page.py
+++ b/src/app_engine/compute_page.py
@@ -22,7 +22,8 @@ ACTION_STATUS = 'status'
 # requeued.
 ACTION_START = 'start'
 
-# Stop the compute instance if it is RUNNING. Then queue a task to start it.
+# Stop the compute instance if it is RUNNING. Then queue a task to start it. Do
+# nothing if the instance is not RUNNING.
 ACTION_RESTART = 'restart'
 
 # Constants for the Compute Engine API
@@ -80,6 +81,8 @@ class ComputePage(webapp2.RequestHandler):
     logging.info('GCE VM \'%s (%s)\' status: \'%s\'.',
                  instance, zone, status)
 
+    # Do nothing if the status is not RUNNING to avoid race. This will cover
+    # most of the cases.
     if status == COMPUTE_STATUS_RUNNING:
       logging.info('Stopping GCE VM: %s (%s)', instance, zone)
       self.compute_service.instances().stop(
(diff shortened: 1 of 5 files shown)


src/app_engine/probers.py
```diff
@@ -49,7 +51,14 @@ def has_non_empty_array_value(dictionary, key):
           dictionary[key])
 
 
+def get_collider_probe_success_key(instance_host):
+  """Returns the memcache key for the last collider instance probing result."""
+  return 'last_collider_probe_success_' + instance_host
+
+
 class ProbeCEODPage(webapp2.RequestHandler):
+  """Page to proble CEOD server."""
```
chuckhays: Typo -> proble/probe

src/app_engine/probers.py
```diff
@@ -155,23 +187,23 @@ def create_collider_active_host(self, old_active_host, probing_results):
...
 
     error_message = None
     result = None
     try:
       result = urlfetch.fetch(
           url=url, method=urlfetch.GET, deadline=PROBER_FETCH_DEADLINE)
-    except Exception as e:
-      error_message = ('urlfetch throws exception: %s, url = %s'
-                       % (str(e), url))
+    except urlfetch.DeadlineExceededError as e:
```
chuckhays: Is this the only exception we might get? Seems like there could be others we care about, certificate errors maybe?

src/app_engine/compute_page.py
```diff
@@ -50,9 +63,30 @@ def _build_compute_service(self):
...
+    if self.compute_service is None:
+      logging.warning('Compute service unavailable.')
+      return
+
+    status = self._compute_status(instance, zone)
+
+    logging.info('GCE VM \'%s (%s)\' status: \'%s\'.',
+                 instance, zone, status)
+
+    if status == COMPUTE_STATUS_RUNNING:
```
chuckhays: What if it is in one of the other states mentioned in the start method?
      # Instance is in an intermediate state: PROVISIONING, STAGING,
      # STOPPING. Thus, requeue the task.
Can we get into a state where we start an extra instead of restarting?

jiayliu:
PTAL

src/app_engine/probers.py
```diff
@@ -155,23 +187,23 @@ def create_collider_active_host(self, old_active_host, probing_results):
 
   def get(self):
     results = {}
-    for collider_instance in constants.WSS_HOST_PORT_PAIRS:
-      results[collider_instance] = self.probe_collider_instance(
-          collider_instance)
+    for instance in constants.WSS_INSTANCES:
```
milkypostman: Why did this change happen here? it doesn't seem to make sense in this context. I mean, the change makes sense separately.

src/app_engine/probers.py
```diff
@@ -104,20 +112,44 @@ def get(self):
 
 
 class ProbeColliderPage(webapp2.RequestHandler):
+  """Page to probe Collider instances."""
+
   def handle_collider_response(
       self, error_message, status_code, collider_instance):
```
milkypostman: I had a hard time following what everything in this file is doing without really going line by line. Could you add a docstring to this function to describe what is happening?

src/app_engine/probers.py
milkypostman: looks like a merge problem.

chuckhays: To store more info for each instance than the hostname. Now it stores the zone as well.

chuckhays:
LGTM barring any updates you want to make for @milkypostman 's comments.

jiayliu:
Added docstring and modified comments per Donald's comments. PTAL

src/app_engine/probers.py
milkypostman: I get it now. It's very confusing since we're keying off host and one case we're keying off something else. I dunno.

jiayliu: I don't understand your confusion. What do you suggest?

On Thu, Mar 5, 2015 at 2:33 PM, Donald Ephraim Curtis <

> In src/app_engine/probers.py
> <https://github.com/webrtc/apprtc/pull/108#discussion_r25908743>:
>
> > @@ -155,23 +187,23 @@ def create_collider_active_host(self, old_active_host, probing_results):
> >
> >    def get(self):
> >      results = {}
> > -    for collider_instance in constants.WSS_HOST_PORT_PAIRS:
> > -      results[collider_instance] = self.probe_collider_instance(
> > -          collider_instance)
> > +    for instance in constants.WSS_INSTANCES:
>
> I get it now. It's very confusing since we're keying off host and one case
> we're keying off something else. I dunno.
>
> —
> <https://github.com/webrtc/apprtc/pull/108/files#r25908743>.
>



-- 
Jiayang Liu

milkypostman:
LGTM w/ comments.

jiayliu:
Changed compute_page from test result. PTAL

milkypostman:
LGTM

chuckhays:
LGTM

PR #109, opened 2015-03-05
Pass metadata from client on call request

chuckhays:
@jiayliu 

Is there more that we wanted to do here or does this cover what we need?


jiayliu:
lgtm :)

milkypostman:
My request would be that in `gcm_notify.py` we document the arguments on *at least* one function **or** we document the `PARAM_METADATA` so that someone can determine what it contains. I was intrigued by the title of this pull request and then I couldn't figure out what metadata was used for at first.

chuckhays:
Added documentation to the code. I had documented it only in the design doc.

PR #110, opened 2015-03-06
Allow branch and commit on build command

chuckhays:
@tkchin 
This goes along with the other review.


chuckhays:
Going a different direction with this.

Issue #118, opened 2015-03-16
JSON.stringify ignores DOM attributes on prototype chains, so you cannot encode DOM attributes.


KaptenJansson:
Spawned from bug: https://code.google.com/p/chromium/issues/detail?id=467471

TL;DR JSON.stringify in Chrome (blink) is more strict as of  https://chromium.googlesource.com/chromium/blink/+/d90fa7c5f4b83a3325b72fed61e8a1af4b0f8c41.

KaptenJansson:
Found the reason, the "this.params_.messages" array from call.js is missing the SDP dict, only candidates are present.

I traced "messages" all the way back to the xhr.responseText in function sendUrlRequest in util.js and it's empty. Maybe add some checking of the "messages" string to make sure to throw an error if it's empty?

<snippet>
XMLHttpRequest {response: "{"result": "SUCCESS", "params": {"pc_constraints":…branch\": \"master\"}", "client_id": "81911118"}}", responseText: "{"result": "SUCCESS", "params": {"pc_constraints":…branch\": \"master\"}", "client_id": "81911118"}}", onreadystatechange: function, timeout: 0, readyState: 3…}onabort: nullonerror: nullonload: nullonloadend: nullonloadstart: nullonprogress: nullonreadystatechange: function () {ontimeout: nullreadyState: 4response: "{"result": "SUCCESS", "params": {"pc_constraints": "{\"optional\": []}", "turn_transports": "", "messages": ["{}", "{\"type\":\"candidate\",\"label\":0,\"id\":\"audio\",\"candidate\":\"candidate:202219810 2 udp 2122262783 2620:0:1043:1:4987:b256:35a8:c8b0 40850 typ host generation 0\"}", "{\"type\":\"candidate\",\"label\":1,\"id\":\"video\",\"candidate\":\"candidate:202219810 2 udp 2122262783
<snippet end> 

I do not have more time to investigate this further but it looks like the array above is empty because the POST self.request.body does not contain the SDP, only the candidates. Something in the POST request is not being parsed correctly in call.js.

line 476 in  apprtc.py
def post(self, room_id, client_id):
    message_json = self.request.body. <<<< Only contains candidates.

KaptenJansson:
So most likely JSON.stringify is failing silently somewhere.

chuckhays:
JSON.stringify in sendSignalingMessage_ was failing to stringify the object containing the SDP info, so it would send {}.

PR #119, opened 2015-03-16
Fix JSON stringify issue with DOM objects

chuckhays:
https://github.com/webrtc/apprtc/issues/118
and
https://code.google.com/p/chromium/issues/detail?id=467471#makechanges

Uses the fix proposed in the chromium bug.

@milkypostman @jiayliu @tkchin

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -38,7 +38,7 @@ module.exports = function(grunt) {
     },
 
     jscs: {
-      src: 'src/**/*.js',
+      src: ['src/**/*.js', '!src/**/enums.js'],
       options: {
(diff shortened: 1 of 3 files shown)


tkchin:
Where are we stringify-ing a DOM object? Afaik we are only stringify-ing dictionaries.
Does this fix the issue though? If so probably stringify-ing returned SDP is the issue.

chuckhays:
We stringify an RTCSessionDescription object in setLocalSdpAndNotify_. RTCSessionDescription appears to be a type that exhibits the problem.

tkchin:
Ok. Can we call it just stringifyObject then since we're not actually using it on the DOM? Also add a comment that it's needed for SDP in particular.

Otherwise lgtm.

Issue #122, opened 2015-03-18
AppRTC broken for Firefox


phoglund:
The AppRTC interop test found that we can no longer bring up calls with a Firefox participant. When I run the test, the Firefox participant gets

'toJSON' called on an object that does not implement interface mozRTCSessionDescription.

on the last line of

function stringifyObject(object) {
  function deepCopy(src) {
    if (typeof src !== "object") {
      return src;
    }
    var dst = Array.isArray(src) ? [] : {};
    for (var property in src) {
      dst[property] = deepCopy(src[property]);
    }
    return dst;
  }
  return JSON.stringify(deepCopy(object));
}

I tried reproing with just connecting two Firefox tabs to apprtc.appspot.com; I couldn't see the error there but the call still didn't get up.

phoglund:
Jiayang: please help with an owner.

You can run the test with out/Debug/browser_tests --gtest_filter="WebRtcApprtc*" --run-manual --gtest_also_run_disabled_tests (follow any instructions the test gives you), but it needs to download a bunch of things so I would not recommend that. You should be able to investigate this with just an apprtc instance and connecting with Firefox.

Also note the Firefox used in the test is a Nightly, so if you can't repro, try with the latest nightly Firefox.

chuckhays:
This looks like a result of the fix I made for Chrome Canary on Monday, I'll take a look.

chuckhays:
What is happening here is that mozRTCSessionDescription includes a toJSON method, which gets copied to the new object when the deep copy is made. When it gets executed by JSON.stringify, the implementation complains that the object isn't the right type.

chuckhays:
Fix in master, should be fixed once deployed.

PR #123, opened 2015-03-18
Stringify for RTCSessionDescription

chuckhays:
This reverts the original fix which worked for Chrome but not Firefox, and makes a simpler change which should work both places.

https://github.com/webrtc/apprtc/issues/122

Firefox's mozRTCSessionDescription includes a toJSON function, probably to solve this exact problem, when the object was copied, the toJSON function was copied as well, and then JSON.stringify called that function. The function checks the object type and throws an error when it isn't the correct type.

Rather than put in more special cases, this change creates the message to send by making a simple copy of the properties we care about.

We could also investigate getting a toJSON added to the Chrome implementation.

@jiayliu @tkchin

### DIFF
--- a/src/web_app/js/call.js
+++ b/src/web_app/js/call.js
@@ -10,7 +10,7 @@
 
 /* globals trace, requestTurnServers, sendUrlRequest, sendAsyncUrlRequest,
    requestUserMedia, SignalingChannel, PeerConnectionClient, setupLoopback,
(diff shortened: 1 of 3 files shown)


jiayliu:
lgtm

KaptenJansson:
Also note hta@'s comment in https://code.google.com/p/chromium/issues/detail?id=468180#c9.

PR #127, opened 2015-03-19
Fix error handling in gcmrecord

jiayliu:
@chuckhays PTAL


jiayliu:
Cover one more case and test added. PTAL @chuckhays

chuckhays:
LGTM

tkchin:
lgtm

PR #131, opened 2015-03-26
change random button label

fippo:
not sure if you have any corporate design guidelines on random buttons... :-)


KaptenJansson:
Not sure about that, @jiayliu do you know? ;)

jiayliu:
I personally think "random" is more clear.

KaptenJansson:
Then we leave it as is.

PR #133, opened 2015-03-26
fixes lint errors in new jscs:srcs

milkypostman:
looks like jscs was updated and changed some of the format to be more strict.

@jiayliu @chuckhays


jiayliu:
lgtm

PR #134, opened 2015-03-26
Store registration id for each verified registration.

tkchin:
@jiayliu @chuckhays PTAL

### DIFF
--- a/src/app_engine/decline_page.py
+++ b/src/app_engine/decline_page.py
@@ -1,5 +1,3 @@
-#!/usr/bin/python2.7
-#
 # Copyright 2015 Google Inc. All Rights Reserved.
 
 """AppRTC Decline Call Handler.
@@ -8,11 +6,12 @@ This module implements declining a call.
 """
 
 import json
-
+import logging
 import webapp2
 
 import constants
 import gcm_notify
+from gcm_notify import GCMByeMessage
 from gcmrecord import GCMRecord
 import room
 import util
@@ -52,19 +51,34 @@ class DeclinePage(webapp2.RequestHandler):
       self.write_response(constants.RESPONSE_INVALID_CALLEE)
       return
 
-    # Notify caller and other callee endpoints about the call decline.
+    # Retrieve caller record and send notifications.
+    caller_record = GCMRecord.get_by_gcm_id(caller_gcm_id, False)
+    gcm_messages = []
+    reason = gcm_notify.GCM_MESSAGE_REASON_TYPE_DECLINED
     # Metadata is passed from callee to caller and other endpoints. Used to
     # indicate why the call was declined, such as being in another call.
     metadata = msg.get(constants.PARAM_METADATA)
-    gcm_ids_to_notify = [caller_gcm_id]
+    if not caller_record:
+      logging.error('Unable to find caller record for gcm_id: %s',
+                    caller_gcm_id)
+      # We still want to attempt to notify callees despite bad state.
+      result = constants.RESPONSE_INTERNAL_ERROR
+    else:
+      # Notify caller about the call decline.
+      gcm_messages.append(GCMByeMessage(caller_gcm_id,
+                                        caller_record.registration_id,
+                                        room_id,
+                                        reason,
+                                        metadata))
+    # Notify other callee endpoints about the call decline.
     for record in callee_records:
       if record.gcm_id != callee_gcm_id:
-        gcm_ids_to_notify.append(record.gcm_id)
-    gcm_notify.send_byes(
-        gcm_ids_to_notify,
-        room_id,
-        gcm_notify.GCM_MESSAGE_REASON_TYPE_DECLINED,
-        metadata)
+        gcm_messages.append(GCMByeMessage(record.gcm_id,
+                                          record.registration_id,
+                                          room_id,
+                                          reason,
+                                          metadata))
+    gcm_notify.send_gcm_messages(gcm_messages)
 
     self.write_response(result)
 
--- a/src/app_engine/gcm_notify.py
+++ b/src/app_engine/gcm_notify.py
@@ -17,6 +17,7 @@ GCM_API_KEY = None
 GCM_API_KEY_PATH = 'gcm_api_key'
 GCM_API_URL = 'https://android.googleapis.com/gcm/send'
 
+GCM_MESSAGE_REGISTRATION_ID_KEY = 'registrationId'
 GCM_MESSAGE_TYPE_BYE = 'BYE'
 GCM_MESSAGE_TYPE_INVITE = constants.ACK_TYPE_INVITE
 GCM_MESSAGE_TYPE_RINGING = 'RINGING'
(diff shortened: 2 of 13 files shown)


src/app_engine/decline_page.py
```diff
@@ -51,13 +51,24 @@ def post(self, room_id):
       self.write_response(constants.RESPONSE_INVALID_CALLEE)
       return
 
+    # Retrieve caller record.
+    caller_record = GCMRecord.get_by_gcm_id(caller_gcm_id, False)
+    if not caller_record:
+      logging.error('Unable to find caller record for gcm_id: %s',
+                    caller_gcm_id)
+      self.write_response(constants.RESPONSE_INTERNAL_ERROR)
+      return
```
jiayliu: I think we still want to notify the callee_records to stop ringing.

src/app_engine/gcm_notify.py
```diff
@@ -44,94 +45,119 @@ def get_gcm_api_key():
...
+  def __init__(self, gcm_id, registration_id):
+    self.gcm_id = gcm_id
+    self.registration_id = registration_id
 
-  Makes a synchronous post to the GCM server with the appropriate payload
-  format.
+  def get_gcm_payload(self):
+    payload_dict = {
+        'data': self.get_data(),
+        'registration_ids': [self.gcm_id],
```
jiayliu: This is now confusing with the registration_id. 

A bigger question is: if the client knows the dest gcm id, why do we need another registration id...
@wzh--

src/app_engine/gcm_notify.py
```diff
@@ -44,94 +45,119 @@ def get_gcm_api_key():
...
+
+  def get_collapse_key(self):
+    return self.room_id
+
+
+class GCMByeMessage(GCMMessage):
+  def __init__(self, gcm_id, registration_id, room_id, reason):
+    super(GCMByeMessage, self).__init__(gcm_id, registration_id)
+    self.gcm_id = gcm_id
+    self.registration_id = registration_id
```
jiayliu: duplicated gcm_id and registration_id

tkchin: Not the same. Registration ID is assigned by AppRTC server per registration. GCM ID is returned by GCM library.

src/app_engine/gcm_notify.py
tkchin: Client doesn't know destination GCM ID. We probably don't want to stuff that in our GCM messages since it is sensitive.

src/app_engine/decline_page.py
```diff
@@ -51,13 +51,26 @@ def post(self, room_id):
       self.write_response(constants.RESPONSE_INVALID_CALLEE)
       return
 
+    # Retrieve caller record.
+    caller_record = GCMRecord.get_by_gcm_id(caller_gcm_id, False)
     # Notify caller and other callee endpoints about the call decline.
-    gcm_ids_to_notify = [caller_gcm_id]
+    gcm_messages = []
```
milkypostman: It would be helpful to split the comment and put `notify caller about decline` here and add `notify callee endpoints about decline` below (before the `for` loop).

src/app_engine/decline_page.py
```diff
@@ -51,13 +51,26 @@ def post(self, room_id):
...
+                                        room_id,
+                                        reason))
     for record in callee_records:
       if record.gcm_id != callee_gcm_id:
-        gcm_ids_to_notify.append(record.gcm_id)
-    gcm_notify.send_byes(
-        gcm_ids_to_notify, room_id, gcm_notify.GCM_MESSAGE_REASON_TYPE_DECLINED)
+        message = GCMByeMessage(
+            record.gcm_id, record.registration_id, room_id, reason)
+        gcm_messages.append(message)
```
milkypostman: nit: no reason to have the message variable here.

src/app_engine/gcm_notify.py
```diff
@@ -44,94 +45,119 @@ def get_gcm_api_key():
...
-  }
-
-
-def send_byes(gcm_ids, room_id, reason):
-  message = create_bye_message(room_id, reason)
-  return send_gcm_messages(gcm_ids, message, room_id)
+  for message in messages:
+    payload = message.get_gcm_payload()
+    rpc = urlfetch.create_rpc()
+    rpc.callback = lambda: handle_rpc_result(rpc, payload)
```
milkypostman: i think i'm ok with lambda but prefer [functools.partial](https://docs.python.org/2/library/functools.html#functools.partial) fro this sort of thing. in this case it doesn't matter but i think with list comprehensions lambda has a lot of overhead. maybe this is not true anymore.

src/app_engine/gcm_notify.py
jiayliu: ... I mean the base class already has these two fields

src/app_engine/gcm_notify.py
jiayliu: Then why are we sending registration_ids: [self.gcm_id] to the client?

src/app_engine/gcm_notify.py
tkchin: Oops. Sorry! No offense intended, lack of coffee yesterday :) Corrected.

src/app_engine/gcm_notify.py
tkchin: Haha, I'm biased because my first real language was LISP. Leaving it since you're okay with it.

src/app_engine/gcm_notify.py
tkchin: Oh, that's the payload for the GCM server to indicate destination. On client side they only see the data field.

jiayliu: I see. Could you add comment to clarify?

tkchin:
PTAL

src/app_engine/gcm_register.py
```diff
@@ -23,7 +21,25 @@
 PARAM_NEW_GCM_ID = 'newGcmId'
 PARAM_USER_ID_LIST = 'userIdList'
 
+REGISTRATION_ID_KEY = 'registrationId'
```
jiayliu: nit: Please add comments to describe the purpose of registration id.

src/app_engine/gcmrecord.py
```diff
@@ -97,7 +102,9 @@ def parse_json(cls, json_string):
       content_dictionary = json.loads(json_string)
       required_keys = [
           cls.USER_ID_KEY, cls.GCM_ID_KEY, cls.CODE_KEY, cls.CODE_SENT_TIME_KEY,
-          cls.VERIFIED_KEY, cls.LAST_MODIFIED_TIME_KEY
+          cls.VERIFIED_KEY, cls.LAST_MODIFIED_TIME_KEY,
+          # TODO(tkchin): add registration id once we clear the store.
+          # cls.REGISTRATION_ID_KEY,
```
jiayliu: Should we clear the store on pushing this change? Old registrations without the new field will fail to receive calls once the client updates.

tkchin: Client will need to be updated before this change is pushed because it changes return signature for some APIs to return JSON instead of just result code. Once that happens we can push this change. Old registrations will still receive the call, just that the regid in the gcm message will be null. We should force re-registration at that point by clearing store so that the receive unintended call bug will be fixed.

jiayliu: SG. 
lgtm

PR #136, opened 2015-03-30
Allow ack and busy messages

chuckhays:
Adds an /ack endpoint for callees to signal to the caller that they have started ringing. A GCM message will be sent to the caller that indicates the call is ringing on the client. The caller will get more than one of these messages if there are multiple ringing endpoints.

/ack -> {type:ringing, calleeGcmId: id, roomId: roomName}

Also adds a metadata field to /decline (same idea as on /join) that is passed in the gcm messages that result. We can use this to auto-decline calls when the user is already in a call and include a reason back to the caller.

@jiayliu @tkchin


src/app_engine/constants.py
```diff
@@ -74,6 +74,9 @@
 PARAM_MESSAGE = 'message'
 PARAM_ROOM_ID = 'roomId'
 PARAM_METADATA = 'metadata'
+PARAM_TYPE = 'type'
+
+TYPE_RINGING = 'ringing'
```
jiayliu: If we go with /ack, I think the 'type' should be the GCM message type that is ack'd.

src/app_engine/join_page.py
```diff
@@ -200,3 +200,90 @@ def post(self, room_id):
         result['is_initiator'])
     logging.info('User ' + client_id + ' joined room ' + room_id +
                  ' with state ' + result['room_state'])
+
+
+class AckPage(webapp2.RequestHandler):
```
jiayliu: Move to its own file?

src/app_engine/join_page.py
```diff
@@ -200,3 +200,90 @@ def post(self, room_id):
...
+    if len(callee_records) < 1:
+      self.report_error(constants.RESPONSE_INVALID_CALLEE)
+      return
+    callee_id = callee_records[0].user_id
+
+    # Notify caller that callee has acknowledged the call and is ringing.
+    # There should only be 1 client in the room, the caller.
+    room_state = call_room.get_room_state()
+    if room_state != room.Room.STATE_WAITING:
+      logging.warning('Room ' + room_id + ' is not in WAITING state ' +
```
jiayliu: Should this be a warning? It's expected to happen for multiple devices.

src/app_engine/join_page.py
```diff
@@ -200,3 +200,90 @@ def post(self, room_id):
...
+
+    if not call_room.is_client_allowed(callee_gcm_id):
+      logging.warning('Room ' + room_id + ' does not allow callee id: ' +
+                      callee_id + '(' + callee_gcm_id + ')')
+      self.report_error(constants.RESPONSE_INVALID_CALLEE)
+      return
+
+    caller_gcm_id = call_room.get_other_client_id(callee_gcm_id)
+    if not caller_gcm_id:
+      logging.warning('Could not find caller gcm id when trying to'
```
jiayliu: Should this be error?

src/app_engine/join_page_test.py
```diff
@@ -224,5 +225,124 @@ def testJoinDifferentRoomTypes(self):
     self.requestCallAndVerify(room_id, 'caller1gcm1', 'callee1',
                               constants.RESPONSE_INVALID_ROOM)
 
+  def testAckRingingInvalidRoom(self):
```
jiayliu: Nice tests!

src/app_engine/join_page_test.py
```diff
@@ -224,5 +225,124 @@ def testJoinDifferentRoomTypes(self):
...
+  def testAckRingingRoomFull(self):
+    self.addTestData()
+    room_id = 'callercallee'
+    self.requestCallAndVerify(room_id, 'caller1gcm1', 'callee1',
+                              constants.RESPONSE_SUCCESS)
+    self.requestAcceptAndVerify(room_id, 'callee1gcm1',
+                                constants.RESPONSE_SUCCESS)
+
+    # Room already full.
+    body = {
```
jiayliu: nit: put creating body, send /ack, and verify error code into a helper method.

src/app_engine/join_page.py
chuckhays: Hmm, I didn't think about the timing issue here. Call from A -> B1/B2, B1 rings immediately and accepts, B2 doesn't get GCM until after accept and the acks the ring, we would hit this case. I'll rethink this. It is important because it is how we get the gcmid of the caller.

jiayliu: Maybe just ignore the request if the room state is not waiting?

chuckhays: That probably makes sense, the request is only useful when in the waiting state.

src/app_engine/constants.py
chuckhays: Makes sense, especially if we add more ack messages. Changed to 'INVITE' to match the gcm message.

src/app_engine/join_page.py
chuckhays: I'm normally the first person to move things to new files, but the code in this is so closely related to the join code that I thought it made sense to keep it here. That said, if we add more ack messages, it won't make sense to have it here. Which way do you think we should go?

src/app_engine/join_page_test.py
chuckhays: Will do. I never think I'll have enough tests to justify doing it, and then all of a sudden I have a bunch of tests that look similar. I should learn my lesson and do it to start with.

chuckhays:
Updated with CR feedback, rebased on apprtc-t which includes jscs fixes.

@jiayliu PTAL

jiayliu:
lgtm

PR #140, opened 2015-04-02
Client_type is now reported as a string.

### LINKED ISSUE #139: client type not properly converted to string
the client_type in the analytics module is not converted to a string. this causes the bigquery request to fail.

milkypostman:
client_Type is passed to the analytics module as a number. It needs to
be converted to a string to be a valid BigQuery entry.

Adds some additional checking in the Analytics unittest code to validate
BigQuery types in the tests. This should catch this type of bug in the future.

Fixes #139

### DIFF
--- a/src/app_engine/analytics.py
+++ b/src/app_engine/analytics.py
@@ -10,7 +10,7 @@ import time
 
 sys.path.append(os.path.join(os.path.dirname(__file__), 'third_party'))
 
-from analytics_enums import EventType, LogField
+from analytics_enums import EventType, LogField, ClientType
 import apiauth
 import constants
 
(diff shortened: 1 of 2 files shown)


src/app_engine/analytics.py
```diff
@@ -75,7 +75,8 @@ def report_event(self, event_type, room_id=None, time_ms=None,
       event[LogField.FLOW_ID] = flow_id
 
     if client_type is not None:
-      event[LogField.CLIENT_TYPE] = client_type
+      client_type_name = ClientType.Name.get(client_type, str(client_type))
```
jiayliu: Can we check types of all the input here or in analytics_page? The client may send any type and we should validate.

milkypostman: We can do that but it's a separate issue. The client was sending the correct data in this case.

I think we should push this change, I filed #141 to track the other issue.

jiayliu:
lgtm

Issue #143, opened 2015-04-16
Unable to establish a loopback call in a new room after leaving a loopback call


tednakamura:
Repro steps:
1) Go to https://apprtc.appspot.com/?debug=loopback, click Join
2) Click the Hangup button
3) Click the "New Room" button
4) Enter a new room name (if one isn't already present), then click Join

Issue: I'm stuck in the "waiting for someone to join this room" state

js console shows these errors:
Room server join error: Registration error: FULL
WebSocket register error: Registration error: FULL

Note that depending on the version of Chrome you're using, if in step 3), you instead click on Rejoin, the loopback call will be established correctly, but audio will fail. That's currently tracked in Chrome bug http://crbug.com/477434.

jiayliu:
It's due to not cleaning up the even listeners and joining the room twice. I'll work on a fix.

jiayliu:
@marwahvikas 
Can you push to AppRTC?

PR #144, opened 2015-04-16
Clean up room selection event listeners to fix joining new foom

jiayliu:


jiayliu:
Fixing https://github.com/webrtc/apprtc/issues/143

chuckhays:
LGTM. Travis seems to be broken though.

PR #145, opened 2015-04-17
Fix stylesheet protocol and correct some broken links

tednakamura:
When viewing params.html over https (i.e. https://apprtc.appspot.com/params.html), the hardcoded stylesheet http href triggers a mixed content error/warning, which, at least in Chrome, prevents the stylesheet from loading.


tednakamura:
I just signed the CLA.

PR #152, opened 2015-05-21
Updates to install instructions.

alvestrand:
Install instructions referred to wrong github repositiories.
The method that worked for installing npm had apparently changed.

GAE version numbers are still confused, but now with different numbers.


phoglund:
lgtm

PR #153, opened 2015-05-21
Change prober URL to reduce false positive alerts

jiayliu:
Other projects running a copy of AppRTC code trigger multiple simultaneous pings to the CEOD server and cause many false positive alerts.
We've changed the prober code to check for project ID, but obviously the other projects didn't update timely.

I'm changing the prober URL, so running older code will no longer ping the CEOD server.


jiayliu:
@chuckhays  PTAL

PR #156, opened 2015-05-22
Add a retry for CEOD prober and adjust the cron job time

jiayliu:
To reduce false alerts and avoid hitting the CEOD server at the same time as other projects hosting old apprtc code.


jiayliu:
@chuckhays  PTAL

chuckhays:
LGTM.

PR #162, opened 2015-06-04
Moar promises

fippo:
Lets dogfood on that promises stuff...

I found a really odd bug... createAnswers gets a third constraints argument which doesn't make sense.
@alvestrand is that worth bailing out on in adapter?

### DIFF
--- a/src/web_app/js/peerconnectionclient.js
+++ b/src/web_app/js/peerconnectionclient.js
@@ -90,9 +90,9 @@ PeerConnectionClient.prototype.startAsCaller = function(offerConstraints) {
       offerConstraints, PeerConnectionClient.DEFAULT_SDP_CONSTRAINTS_);
   trace('Sending offer to peer, with constraints: \n\'' +
       JSON.stringify(constraints) + '\'.');
-  this.pc_.createOffer(this.setLocalSdpAndNotify_.bind(this),
-      this.onError_.bind(this, 'createOffer'),
-      constraints);
+  this.pc_.createOffer(constraints)
+  .then(this.setLocalSdpAndNotify_.bind(this))
+  .catch(this.onError_.bind(this, 'createOffer'));
 
   return true;
 };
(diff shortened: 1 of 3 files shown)


src/web_app/js/peerconnectionclient.js
```diff
@@ -174,9 +174,9 @@ PeerConnectionClient.prototype.getPeerConnectionStats = function(callback) {
 
 PeerConnectionClient.prototype.doAnswer_ = function() {
   trace('Sending answer to peer.');
-  this.pc_.createAnswer(this.setLocalSdpAndNotify_.bind(this),
-      this.onError_.bind(this, 'createAnswer'),
-      PeerConnectionClient.DEFAULT_SDP_CONSTRAINTS_);
+  this.pc_.createAnswer()
```
jiayliu: The answer constraints should not be removed, webrtc looks for it to decide if it receives media or uses bundle:

https://code.google.com/p/chromium/codesearch#chromium/src/third_party/libjingle/source/talk/app/webrtc/mediastreamsignaling.cc&rcl=1437050479&l=442

fippo: hrm... that should hopefully work with https://github.com/webrtc/adapter/blob/master/adapter.js#L228-229
now if firefox doesn't error i can fix that.

fippo: Firefox throws an error unfortunately :-(

fippo:
@jiayliu worked around the issue -- ugly but *shrug*

I've also seen getStats failures since apprtc is using the legacy syntax without a browser switch but that's a different story

jiayliu:
Does it mean we have a regression in infobox on firefox?

fippo:
I doubt it's a regression. Firefox getStats has never worked without a track selector as its called in https://github.com/webrtc/apprtc/blob/master/src/web_app/js/peerconnectionclient.js#L172

jiayliu:
lgtm.
ok to merge?

fippo:
yeah. The Firefox issue should soon be fixed but I guess it will take time to go to the stable channel there.

Issue #187, opened 2015-08-11
Manifest "density" parsing error


KaptenJansson:
You can see the error in https://cloud.githubusercontent.com/assets/9001606/9183286/0af238aa-3f63-11e5-9eeb-73f2e3d1be5e.png (from #181).

This is strange, I've seen [examples](https://crosswalk-project.org/documentation/manifest.html) of adding density as a true float ("2.0") and as an int ("2).
https://github.com/webrtc/apprtc/blob/master/src/web_app/html/manifest.json#L9

I wonder why it fails?

KaptenJansson:
@selavazhagan FYI

samdutton:
I think this might be because the value is a string ("4.0") not a float.

Attempted fix in #188.

Can't seem to reproduce this — what Chrome version is the screenshot?

KaptenJansson:
Well it's a string in all the examples, I assumed it still treated it as a float.

The density property seems to be affected by the screen density and it affects CrOS device the most it seems, so if you have a CrOS give it a shot.

samdutton:
Got it: can repro on Chrome OS after upgrading to 46 (but not before).

KaptenJansson:
I updated it and it solved it but not immediately, you might need to clear you browsing history.

Thanks for fixing it.

PR #188, opened 2015-08-11
Attempt to fix manifest parsing error for icon density value #187

samdutton:
As per #187.

Changed density value from string to float.


KaptenJansson:
I can repro on a CrOS device, I will merge this and test it.

samdutton:
See #189: advice is now to remove density setting.

PR #189, opened 2015-08-11
Remove density setting for icon in manifest.json

samdutton:
On advice from Paul Kinlan: icons are chosen depending on resolution. Including a density value may cause an icon not to be chosen when it's actually the best/only option.


KaptenJansson:
LGTM

PR #191, opened 2015-08-31
Use mediaStreamTrack.stop method

KaptenJansson:
This fails due to phantomjs uses webkit Safari 534.34 which is from 2011.

Will make a separate PR for updating/replacing phantomjs.


KaptenJansson:
FRT it stalls on mediaStream.getTracks() (or getVideo/audioTracks)

KaptenJansson:
Fixed in #198

PR #193, opened 2015-09-02
Bump adapter to 0.2.3

### LINKED ISSUE #192: Console logging broken
That is, nothing shows up in the console log when using the current version of apprtc. I think this is because the fix in https://github.com/webrtc/adapter/issues/110 hasn't been downstreamed into the copy of adapter.js in this repo, but I'm not certain.

KaptenJansson:
Fixes #192


KaptenJansson:
Verified it in https://3-dot-apprtc.appspot.com/.

PR #198, opened 2015-09-25
Close tracks rather than the media stream.

phoglund:
mediaStream.stop() is deprecated and has been removed; this makes us close the tracks instead.

Context: https://code.google.com/p/chromium/issues/detail?id=535588.

### DIFF
--- a/src/web_app/js/call.js
+++ b/src/web_app/js/call.js
@@ -118,9 +118,14 @@ Call.prototype.hangup = function(async) {
   }
 
   if (this.localStream_) {
-    this.localStream_.getTracks().forEach(function(track) {
-      track.stop();
-    });
+    if (typeof this.localStream_.stop === "function") {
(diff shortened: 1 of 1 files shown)


phoglund:
Please review. This will break the build, but it's broken anyway (https://github.com/webrtc/apprtc/issues/199).

phoglund:
Found a way to not break the build.

src/web_app/js/call.js
```diff
@@ -118,7 +118,14 @@ Call.prototype.hangup = function(async) {
   }
 
   if (this.localStream_) {
-    this.localStream_.stop();
+    if (typeof this.localStream_.stop === "function") {
```
KaptenJansson: Use single quotes.

phoglund:
PTAL

KaptenJansson:
LGTM

PR #200, opened 2015-09-29
Check if getTracks exists

KaptenJansson:
This makes sure we do not use mediaStream.stop() if it exists on browsers that support getTracks as well.


phoglund:
lgtm

PR #201, opened 2015-10-01
Copy in a way that's supported on Windows.

phoglund:
Your patch that introduced "cp" breaks AppRTC for win development. You basically can't run shell commands - everything needs to be Python.


KaptenJansson:
LGTM

PR #215, opened 2015-11-04
Remove the minVersion config for the websocket server

jiayliu:
It caused failures on older Android devices.


KaptenJansson:
LGTM

PR #220, opened 2015-11-13
Check for undefined codec

KaptenJansson:
If no codec parameters are specified, codec is undefined not an empty string.

I left the condition for an empty string due to I suppose we should handle the case where the vsc & vrc parameters are provided without any codec specified.

### DIFF
--- a/src/web_app/js/sdputils.js
+++ b/src/web_app/js/sdputils.js
@@ -226,7 +226,7 @@ function maybePreferVideoReceiveCodec(sdp, params) {
 // The format of |codec| is 'NAME/RATE', e.g. 'opus/48000'.
 function maybePreferCodec(sdp, type, dir, codec) {
   var str = type + ' ' + dir + ' codec';
-  if (codec === '') {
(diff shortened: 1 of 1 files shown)


jiayliu:
lgtm

src/web_app/js/sdputils.js
```diff
@@ -226,7 +226,7 @@ function maybePreferVideoReceiveCodec(sdp, params) {
 // The format of |codec| is 'NAME/RATE', e.g. 'opus/48000'.
 function maybePreferCodec(sdp, type, dir, codec) {
   var str = type + ' ' + dir + ' codec';
-  if (codec === '') {
+  if (typeof codec === 'undefined' || codec === '') {
```
juberti: Did you consider doing if (!codec), which covers both cases?

KaptenJansson: I did not as I was thinking about the issue explicitly, your suggestion covers all the false cases. Will update.

PR #228, opened 2015-12-07
Add warningMessages function for non fatal warnings, fix encode issue

### LINKED ISSUE #211: SyntaxError when both "hd" and "video" URL parameters are present
Repro steps:
1) Try to start an apprtc call where both the "hd" and "video" URL parameters are set. Example: https://apprtc.appspot.com/?debug=loopback&video=googNoiseReduction=true&hd=true
2) Open the js console

Expected result:
Call starts, though console shows the warning message "The "hd" parameter has overridden video=$video_params"

Actual result:
Call doesn't start, console shows "Uncaught SyntaxError: Unexpected string" 
Clicking that console line entry takes me to a source code listing that looks something like:

`<script type="text/javascript">var loadingParams={errorMessages:[u'The "hd" parameter has overridden video=googNoiseReduction=true'],isLoopback:true,mediaConstraints:{"audio":true,"video":{"optional":[],"mandatory":{"minWidth":"1280","minHeight":"720"}}},offerConstraints:{"optional":[],"mandatory":{}},peerConnectionConfig:{"iceServers":[]},peerConnectionConstraints:{"optional":[{"DtlsSrtpKeyAgreement":false}]},` (remaining text snipped)

KaptenJansson:
Fixes #211 

* Add warning_messages reporting functionality, both the HD + video and minre params do not have to be fatal hence making a separate message functionality for those type of messages.
* The issue in #211 is that the [HTML template](https://github.com/webrtc/apprtc/blob/master/src/web_app/html/index_template.html) is UTF-8 while python uses unicode by default. This results in python appending a `u` for unicode before the string in the array `[u'string']` which does not make Javascript happy ;), it does not affect all messages though, possible due to including the URL parameter in the string.

A more permanent solution would to create a wrapper that converts all incoming strings saving the use of string.encode('utf-8') on all messages to be pushed to a template. But since there were only two warning_messages to be printed and no error_messages I do not want to spend the time on it, also we might use a different templating framework in the future.

### DIFF
--- a/src/app_engine/apprtc.py
+++ b/src/app_engine/apprtc.py
@@ -217,7 +217,7 @@ def get_room_parameters(request, room_id, client_id, is_initiator):
   if hd and video:
     message = 'The "hd" parameter has overridden video=' + video
     logging.warning(message)
-    # HTML template is UTF-8 while python defaults to unicode.
+    # HTML template is UTF-8, make sure the string is UTF-8 as well.
     warning_messages.append(message.encode('utf-8'))
   if hd == 'true':
     video = 'mandatory:minWidth=1280,mandatory:minHeight=720'
@@ -228,7 +228,7 @@ def get_room_parameters(request, room_id, client_id, is_initiator):
     message = ('The "minre" and "maxre" parameters are no longer ' +
         'supported. Use "video" instead.')
     logging.warning(message)
-    # HTML template is UTF-8 while python defaults to unicode.
+    # HTML template is UTF-8, make sure the string is UTF-8 as well.
     warning_messages.append(message.encode('utf-8'))
 
   # Options for controlling various networking features.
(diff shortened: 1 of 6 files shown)


src/app_engine/apprtc.py
```diff
@@ -215,18 +216,20 @@ def get_room_parameters(request, room_id, client_id, is_initiator):
   hd = request.get('hd').lower()
   if hd and video:
     message = 'The "hd" parameter has overridden video=' + video
-    logging.error(message)
-    error_messages.append(message)
+    logging.warning(message)
+    # HTML template is UTF-8 while python defaults to unicode.
```
andresusanopinto: This comment makes no sense. UTF-8 is an encoding of all the chars in unicode standard.

andresusanopinto:
LGTM besides the comment.

PR #230, opened 2015-12-07
Hide audio/video mute buttons if no corresponding track is detected

### LINKED ISSUE #202: Mute/UnMute Audio/Video option enabled for "Video only"/"Audio only" call respectively
Description:
When the user makes an appRTC call with "audio=false&video=true" (or "audio=true&video=false") parameters, he/she sees that the "Mute audio" / "Mute Video" icon can be toggled on/off respectively.

Steps to reproduce:
1. Append "audio=false&video=true" (or "audio=true&video=false") parameters to an appRTC call and click Join.
2. Hover over the "Mute audio" / "Mute Video" icons and toggle them on/off 
Sample Test page - https://apprtc.appspot.com/r/862942399?audio=false&video=true

Actual results:
The user is able to mute/unmute Audio on a "Video only" call (and/or) mute/unmute Video on an "Audio only" call  

Expected results:
- Audio icon should be disabled/invisible for a Video only call
- Video icon should be disabled/invisible for an Audio only call

KaptenJansson:
Fixes #202  

Also enable the new travis environment

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: node_js
 node_js:
   - 0.10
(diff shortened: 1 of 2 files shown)


samdutton:
LGTM.

PR #231, opened 2015-12-08
Update collider_test.go websocket dependency.

hkjellander:
The websocket hosted on code.google.com no longer exists,
which breaks our Chromium bots in
https://build.chromium.org/p/chromium.webrtc/waterfall

BUG=http://crbug.com/567598


hkjellander:
I signed it!

KaptenJansson:
LGTM!

Issue #224, opened 2015-11-27
Make VP9 default video codec


mflodman:
Chrome M48 contains VP9 support without a flag, we should make VP9 the default apprtc video codec for apprtc and browsers supporting VP9.

mflodman:
Current URL parameter to select VP9:
vsc=101 VP9/90000

KaptenJansson:
Hmm I forgot about this, also do we need to check if it's Chrome?

@andresusanopinto was a bit quick reviewing your PR:(

andresusanopinto:
No need to check it is chrome, it simply moves VP9 to be the first on the list if the codec is present. e.g. ?vsc=VPN-does-not-exists-lala" would still work and fallback to browser default (i.e. VP8).

PR #234, opened 2015-12-09
Make VP9 the default video send and receive codec.

andresusanopinto:


KaptenJansson:
LGTM

andresusanopinto:
Fixed #224

src/web_app/js/appcontroller.js
```diff
@@ -519,9 +519,9 @@ AppController.prototype.loadUrlParams_ = function() {
   this.loadingParams_.opusStereo = urlParams['stereo'];
   this.loadingParams_.videoSendBitrate = urlParams['vsbr'];
   this.loadingParams_.videoSendInitialBitrate = urlParams['vsibr'];
-  this.loadingParams_.videoSendCodec = urlParams['vsc'];
+  this.loadingParams_.videoSendCodec = urlParams['vsc'] || 'VP9';
```
juberti: I tend to think that we should only do this for the receive codec. IOW, we should just change the outgoing offer to promote VP9, we shouldn't clobber the incoming offer.

Having a DEFAULT_VIDEO_CODEC constant rather than hardcoding 'VP9' would also be a good idea.

andresusanopinto: Agreed. We should only modify our local sdp.

PR #244, opened 2015-12-15
Improve parameter handling

KaptenJansson:
* Enable autoescape in jinja
* Fix all parsing handling as a result of enabling HTML escaping
* Report TURN error as a warning instead of a status message.


src/web_app/html/index_template.html
```diff
@@ -124,24 +124,23 @@
 {% if not chromeapp %}
   <script type="text/javascript">
     var loadingParams = {
```
andresusanopinto: Have you consideed dumping the whole loading params as json?

src/web_app/js/call.js
```diff
@@ -18,6 +18,40 @@
 
 var Call = function(params) {
   this.params_ = params;
+
```
andresusanopinto: I don't like this much. You should not need to care about escape unescaping here.

src/web_app/html/index_template.html
andresusanopinto: Here I would expect a simple:
 var loadingParams = {{loadingParams}};

In python code i would expect like:
  loadginParams = get_room_parameters(self.request, None, None, None)
  self.write_response('index_template.html', {'loadingParams': json.dumps(loadingParams) })

Also whenever adding new params it would mean one less place to edit the code.

KaptenJansson: I originally did this but I wanted the granularity and be able to chose what is JSON and what's not due to some is already JSON. But yes, this is a much cleaner approach but it either requires cleaning up so that we only encode it once or we have to decode it twice.

I will redo this.

src/web_app/js/call.js
KaptenJansson: I do not want to store them unescaped in the DOM but I will try to escape them earlier.

KaptenJansson:
Closing this in favor of a new PR #250

PR #246, opened 2015-12-18
Use the standard page visibility API

phistuck:
Changed webkitFoo to foo for the page visibility API.


KaptenJansson:
Thanks for the PR.

LGTM.

PR #254, opened 2016-01-26
Correcting collider symlink instructions.

phoglund:
I found after quite a bit of work that the collider build instructions only work if one's directories are arranged in a certain way. This correction should work for everyone.


KaptenJansson:
aha did not know about the -r switch, awesome! LGTM.

PR #259, opened 2016-02-02
Add ECDSA certificate generation + update adapter.js to 0.2.9

### LINKED ISSUE #258: Add support for EC certs using generateCertificates
This should modestly improve startup performance.

When creating the peerconnection, should be able to check for the existence of the generateCertificate API, and if present, generate a new EC cert for the PC to use.

KaptenJansson:
Fixes #258

### DIFF
--- a/src/web_app/js/peerconnectionclient.js
+++ b/src/web_app/js/peerconnectionclient.js
@@ -29,6 +29,22 @@ var PeerConnectionClient = function(params, startTime) {
     '  constraints: \'' + JSON.stringify(params.peerConnectionConstraints) +
     '\'.');
 
+  if (typeof RTCPeerConnection.generateCertificate === 'function') {
+    trace('Trying to generate an ECDSA certificate.');
+    RTCPeerConnection.generateCertificate({name: 'ECDSA', namedCurve: 'P-256'})
+    .then(function(cert) {
+      if (cert.expires) {
+        trace('Certificate generated successfully.');
+        params.peerConnectionConfig.certificates = [cert];
+      } else {
+        trace('Could not generate an ECDSA certificate, using default.');
+      }
+    })
+    .catch(function(error) {
+      trace('Could not generate a certificate: ' + error);
+    });
+  }
+
   // Create an RTCPeerConnection via the polyfill (adapter.js).
   this.pc_ = new RTCPeerConnection(
       params.peerConnectionConfig, params.peerConnectionConstraints);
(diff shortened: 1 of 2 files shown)


src/web_app/js/peerconnectionclient.js
```diff
@@ -29,6 +29,22 @@ var PeerConnectionClient = function(params, startTime) {
...
     '\'.');
 
+  if (typeof RTCPeerConnection.generateCertificate === 'function') {
+    trace('Trying to generate an ECDSA certificate.');
+    RTCPeerConnection.generateCertificate({name: 'ECDSA', namedCurve: 'P-256'})
+    .then(function(cert) {
+      if (cert.expires) {
+        trace('Certificate generated successfully.');
+        params.peerConnectionConfig.certificates = [cert];
+      } else {
```
juberti: This is confusing. Why do we need two different error handlers? Is this an artifact of wrapping generateCertificates?

cc @henbos

KaptenJansson: Yes it might be, during testing generateCertificate was successful but the RTCCertificate object was empty on M49 and did not contain the `expires` property thus I assumed it failed or it might have succeeded and the property is missing. On M50 the `expires` property is always there.

@henbos, do you know anything about the discrepancy as described above?

henbos: In M49 we had not implemented the expires attribute, now we have. There will always be an expires attribute; you can remove the if-else here.

juberti: I think it's fine that the cert object is empty - it's just a handle. 

I don't really know why we care about .expires; it's a self-signed cert, after all.

henbos: That's correct. It looks empty but there's magical C++ stuff not visible in javascript associated with it.

KaptenJansson: I see, then we will rely on the promise and error handler of the API (makes sense;)). Removed the extra error handler.

juberti:
lgtm - ship it

juberti:
please update when this is live

KaptenJansson:
It's now live

src/web_app/js/peerconnectionclient.js
```diff
@@ -29,6 +29,17 @@ var PeerConnectionClient = function(params, startTime) {
     '  constraints: \'' + JSON.stringify(params.peerConnectionConstraints) +
     '\'.');
 
+  if (typeof RTCPeerConnection.generateCertificate === 'function') {
+    RTCPeerConnection.generateCertificate({name: 'ECDSA', namedCurve: 'P-256'})
+    .then(function(cert) {
+      trace('Certificate generated successfully.');
+      params.peerConnectionConfig.certificates = [cert];
```
juberti: Just realized this code doesn't work (AppRTC continues to use RSA certs).

You can't create the peerconnection until after the promise resolves. Might need to create the PC later, perhaps in startAsCaller/Callee.

PR #262, opened 2016-02-05
Add a redirect list

### LINKED ISSUE #221: www.appr.tc, apprtc.net, should redirect to appr.tc
Otherwise, they will have different permissions, cookies, etc.

KaptenJansson:
Fixes #221

### DIFF
--- a/src/app_engine/apprtc.py
+++ b/src/app_engine/apprtc.py
@@ -534,8 +534,10 @@ class MainPage(webapp2.RequestHandler):
 
   def get(self):
     """Renders index.html."""
-    if self.request.headers['Host'] == 'apprtc.net':
-      webapp2.redirect('https://www.apprtc.net', permanent=True)
+    REDIRECT_DOMAINS = ['apprtc.appspot.com', 'apprtc.net', 'www.apprtc.net',
+        'apprtc.webrtc.org', 'www.appr.tc']
+    if self.request.headers['Host'] in REDIRECT_DOMAINS:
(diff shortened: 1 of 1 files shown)


src/app_engine/apprtc.py
```diff
@@ -534,8 +534,10 @@ def write_response(self, target_page, params={}):
 
   def get(self):
     """Renders index.html."""
-    if self.request.headers['Host'] == 'apprtc.net':
-      webapp2.redirect('https://www.apprtc.net', permanent=True)
+    REDIRECT_DOMAINS = ['apprtc.appspot.com', 'apprtc.net', 'www.apprtc.net',
+        'apprtc.webrtc.org', 'www.appr.tc']
+    if self.request.headers['Host'] in REDIRECT_DOMAINS:
```
juberti: do we want to add the path, if one exists? i.e. redirect apprtc.net/r/foo to appr.tc/r/foo

KaptenJansson: The problem is that the URL redirects handled by the ISP do not handle redirects with path. This means the ones we redirect using this hack will behave differently than the ones redirected by the ISP.

juberti: True, but hopefully use of apprtc.net should be rare.

I do think we want to ensure that at least apprtc.appspot.com/r/foo redirects properly.

KaptenJansson: Makes sense, will fix.

PR #264, opened 2016-02-12
Make sure certificate is generated before a peerconnection has been created

### LINKED ISSUE #258: Add support for EC certs using generateCertificates

KaptenJansson:
I had to move the cert creation out of the peerconnectionclient into call.js which is fine IMHO since it's actually creating/adding parameters for the peerConnection which call.js is already doing.

Fixes #258

### DIFF
--- a/src/web_app/js/call.js
+++ b/src/web_app/js/call.js
@@ -426,22 +426,15 @@ Call.prototype.maybeCreatePcClientAsync_ = function() {
       var certParams = {name: 'ECDSA', namedCurve: 'P-256'};
       RTCPeerConnection.generateCertificate(certParams)
       .then(function(cert) {
-        trace('ECDSA Certificate generated successfully.');
+        trace('ECDSA certificate generated successfully.');
         this.params_.peerConnectionConfig.certificates = [cert];
         this.createPcClient_();
         resolve();
       }.bind(this))
       .catch(function(error) {
-        if (this.params_.peerConnectionConfig.certificates) {
-          reject(error);
-        } else {
-          // This is not a critical error hence why it should continue
-          // setting up the call.
-          trace('Could not generate a certificate: ' + error);
-          this.createPcClient_();
-          resolve();
-        }
-      }.bind(this));
+        trace('ECDSA certificate generation failed.');
+        reject(error);
+      });
     } else {
       this.createPcClient_();
       resolve();
--- a/src/web_app/js/call.js
+++ b/src/web_app/js/call.js
@@ -419,23 +419,45 @@ Call.prototype.maybeCreatePcClient_ = function() {
   if (this.pcClient_) {
     return;
   }
-  try {
-    this.pcClient_ = new PeerConnectionClient(this.params_, this.startTime);
-    this.pcClient_.onsignalingmessage = this.sendSignalingMessage_.bind(this);
-    this.pcClient_.onremotehangup = this.onremotehangup;
-    this.pcClient_.onremotesdpset = this.onremotesdpset;
-    this.pcClient_.onremotestreamadded = this.onremotestreamadded;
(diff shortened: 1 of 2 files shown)


src/web_app/js/call.js
```diff
@@ -419,23 +419,45 @@ Call.prototype.maybeCreatePcClient_ = function() {
   if (this.pcClient_) {
     return;
```
juberti: No promise is returned here, which seems like a problem.

src/web_app/js/call.js
```diff
@@ -419,23 +419,45 @@ Call.prototype.maybeCreatePcClient_ = function() {
...
+        trace('ECDSA Certificate generated successfully.');
+        this.params_.peerConnectionConfig.certificates = [cert];
+        this.createPcClient_();
+        resolve();
+      }.bind(this))
+      .catch(function(error) {
+        if (this.params_.peerConnectionConfig.certificates) {
+          reject(error);
+        } else {
+          // This is not a critical error hence why it should continue
```
juberti: When would this happen? This fallback logic makes this function more complex.

src/web_app/js/call.js
```diff
@@ -419,23 +419,45 @@ Call.prototype.maybeCreatePcClient_ = function() {
   if (this.pcClient_) {
```
juberti: This function should probably be called maybeCreatePcClientAsync or some such to differentiate it from the synchronous createPcClient.

src/web_app/js/call.js
KaptenJansson: I did this because I did not want a certificate generation failure (we would fallback to the default certificate right?) to prevent the call from being setup but still catch other errors which most likely are fatal.

src/web_app/js/call.js
KaptenJansson: Well spotted, I think we do reuse when we rejoin the call but have not confirmed this 100%.

src/web_app/js/call.js
juberti: If we fail generating this cert, we probably have a significant bug - which we would want to discover as soon as possible. I'd prefer to fail hard if we can't generate the cert.

src/web_app/js/call.js
juberti: Can you double-check this, and add a comment for the next reader?

src/web_app/js/call.js
KaptenJansson: Ok that makes sense, will remove the extra logic.

src/web_app/js/call.js
KaptenJansson: Done.

src/web_app/js/call.js
KaptenJansson: It hits this check 3 times when another user joins the call (via startSignaling_ which is called on in getTurnServicersPromise_ and onRemoteHangup. Also via onRecvSignalingChannelMessage_ which is hooked up to the this.channel_.onmessage event) but not before. Also when a user (B) leaves and rejoins, the user (A) left in the call hits this 2 check times while the user (B) rejoins.

What comment would be appropriate here? E.g.
```
// Resolve promise if pcClient already exists.
```

juberti: Does it get called multiple times in a single call? 

I can understand the case where B leaves and rejoins, A uses the same PC, but it's not clear to me why this would be hit multiple times in a single call.

juberti: Looked more closely, and I agree with your analysis. I don't think there's a simple comment that really adds a lot of value, so I think it's fine as-is.

juberti:
lgtm

KaptenJansson:
And it's now pushed as well.

PR #270, opened 2016-03-01
Add new turn service

### LINKED ISSUE #219: Switch AppRTC to new TURN service
Need to make sure this is api keyed appropriately to limit this only to AppRTC use.

cc @samdutton @KaptenJansson

KaptenJansson:
**Description**
Fixes #219.

* Disabled cron job for CEOD proper but leave the prober code intact for now.
* Uses app.yml environment variable for the API key and can be replaced when deploying via `appcfg.py ---env_variable=TURN_API_KEY:KEY`

**Purpose**
Increase the reliability and success rate of AppRTC calls.

@juberti cc


KaptenJansson:
I will add this API key handling to the callstats PR as well, much neater.

src/web_app/js/util.js
```diff
@@ -89,17 +85,12 @@ function requestTurnServers(turnRequestUrl, turnTransports) {
       }
       // Filter the TURN URLs to only use the desired transport, if specified.
       if (turnTransports.length > 0) {
-        filterTurnUrls(turnServerResponse.uris, turnTransports);
+        for (var i = 0; i < turnServerResponse.iceServers.length; i++) {
+          filterTurnUrls(turnServerResponse.iceServers[i].urls, turnTransports);
```
andresusanopinto: Remember to drop the ice servers that end up wuth an empty urls. Maybe copy the filterIceConfig method from testrtc.

src/app_engine/constants.py
```diff
@@ -11,9 +11,9 @@
 
 LOOPBACK_CLIENT_ID = 'LOOPBACK_CLIENT_ID'
 
-TURN_BASE_URL = 'https://computeengineondemand.appspot.com'
-TURN_URL_TEMPLATE = '%s/turn?username=%s&key=%s'
-CEOD_KEY = '4080218913'
+TURN_BASE_URL = 'https://networktraversal.googleapis.com'
```
andresusanopinto: I think having a single TURN_URL = 'https://networktraversal.googleapis.com/v1alpha/iceconfig?key=' + os.environ would be simpler than having these 3 variables.

src/web_app/js/util.js
KaptenJansson: Right, did not think it would here as we use our turn server but people can ofc specify a turn server in the URL parameters or by using their own turn server provider. Sorry for not bringing over the lessons learnt.

src/app_engine/constants.py
KaptenJansson: Sure.

KaptenJansson: Actually it follows the http://tools.ietf.org/html/draft-uberti-rtcweb-turn-rest-00#section-2 draft. I know the new turn service do not but maybe it's easier to keep it close to what others might use?

andresusanopinto: sgtm

src/web_app/js/util.js
```diff
@@ -74,32 +74,21 @@ function sendUrlRequest(method, url, async, body) {
   });
 }
 
-// Returns a list of turn servers after requesting it from CEOD.
+// Returns a list of turn servers after requesting it from the TURN server.
```
juberti: from... GCP? Surely we don't request the turn server from the turn server

src/web_app/js/util.js
```diff
@@ -74,32 +74,21 @@ function sendUrlRequest(method, url, async, body) {
...
-    var method = isChromeApp() ? 'POST' : 'GET';
-    sendAsyncUrlRequest(method, turnRequestUrl).then(function(response) {
+    sendAsyncUrlRequest('POST', turnRequestUrl).then(function(response) {
       var turnServerResponse = parseJSON(response);
       if (!turnServerResponse) {
         reject(Error('Error parsing response JSON: ' + response));
         return;
       }
-      // Filter the TURN URLs to only use the desired transport, if specified.
+      turnTransports = 'tcp';
```
juberti: why are we setting this here?

src/web_app/js/util.js
```diff
@@ -74,32 +74,21 @@ function sendUrlRequest(method, url, async, body) {
   });
 }
 
-// Returns a list of turn servers after requesting it from CEOD.
+// Returns a list of turn servers after requesting it from the TURN server.
 function requestTurnServers(turnRequestUrl, turnTransports) {
```
juberti: It might be useful to put a sample response in a comment above requestTurnServers.

src/web_app/js/util.js
```diff
@@ -118,15 +107,27 @@ function parseJSON(json) {
...
 
 // Filter a list of TURN urls to only contain those with transport=|protocol|.
-function filterTurnUrls(urls, protocol) {
-  for (var i = 0; i < urls.length;) {
-    var parts = urls[i].split('?');
-    if (parts.length > 1 && parts[1] !== ('transport=' + protocol)) {
-      urls.splice(i, 1);
-    } else {
-      ++i;
+function filterTurnUrls(config, protocol) {
```
juberti: Can we just filter each iceServer in the turnServer response? Not sure why the function had to be rewritten.

src/web_app/js/util.js
```diff
@@ -74,32 +74,21 @@ function sendUrlRequest(method, url, async, body) {
...
-    sendAsyncUrlRequest(method, turnRequestUrl).then(function(response) {
+    sendAsyncUrlRequest('POST', turnRequestUrl).then(function(response) {
       var turnServerResponse = parseJSON(response);
       if (!turnServerResponse) {
         reject(Error('Error parsing response JSON: ' + response));
         return;
       }
-      // Filter the TURN URLs to only use the desired transport, if specified.
+      turnTransports = 'tcp';
       if (turnTransports.length > 0) {
```
andresusanopinto: Nit: use turnTransports !=== "". Using length made me think it was an array.

src/web_app/js/util.js
andresusanopinto: This function leaves the stun servers if they specify the requested protocol, but does not modifies an unspecified stun address to add the protocol. It feels incoherent to me.

Suggestion 1: rename to filterIceServersUrls and treat all urls the same independently of startsWith.
Suggestion 2: filter out all urls that do not startwith('turn') even if they specific the right transport.

The reason why this needed to be re-written is that this code was creating an iceServer entry with no urls since networktraversal reports stun server aside from turn servers.

src/web_app/js/util.js
KaptenJansson: My bad, used for testing the filter mechanism. Will remove.

src/web_app/js/util.js
KaptenJansson: Haha, well spotted, should have been Turn server provider.

src/web_app/js/util.js
KaptenJansson: SGTM

src/web_app/js/util.js
KaptenJansson: Well spotted! Will def change it.

src/web_app/js/util.js
KaptenJansson: Actually ICE server provider is better IMHO.

src/web_app/js/util.js
KaptenJansson: I went with suggestion 1. Also renamed uri to url.

src/web_app/js/util.js
```diff
@@ -74,32 +74,36 @@ function sendUrlRequest(method, url, async, body) {
...
+// {
+//   lifetimeDuration: "43200.000s",
+//   iceServers: [
+//     {
+//       urls: ["turn:1.2.3.4:19305", "turn:1.2.3.5:19305"],
+//       username: "username",
+//       credential:"credential"
+//     },
+//     {
+//       urls: ["stun:stun.l.google.com:19302"]
```
andresusanopinto: Just place a random ip on this one too. Or "stun.example.com"

src/web_app/js/util.js
```diff
@@ -118,15 +122,27 @@ function parseJSON(json) {
 }
 
 // Filter a list of TURN urls to only contain those with transport=|protocol|.
```
andresusanopinto: Filter a peer connection config to only contain...

src/web_app/js/utils_test.js
```diff
@@ -32,15 +32,15 @@ var TURN_URLS_TCP = [
 
 var UtilsTest = new TestCase('UtilsTest');
 
-UtilsTest.prototype.testFilterTurnUrlsUdp = function() {
+UtilsTest.prototype.filterIceServersUrlsUdp = function() {
   var urls = TURN_URLS.slice(0);  // make a copy
-  filterTurnUrls(urls, 'udp');
+  filterIceServersUrls(urls, 'udp');
```
andresusanopinto: This method works over a peer connection config. Not over an array of urls.

src/web_app/js/utils_test.js
```diff
@@ -32,15 +32,15 @@ var TURN_URLS_TCP = [
...
   assertEquals('Only transport=udp URLs should remain.', TURN_URLS_UDP, urls);
 };
 
-UtilsTest.prototype.testFilterTurnUrlsTcp = function() {
+UtilsTest.prototype.filterIceServersUrlsTcp = function() {
   var urls = TURN_URLS.slice(0);  // make a copy
-  filterTurnUrls(urls, 'tcp');
+  filterIceServersUrls(urls, 'tcp');
```
andresusanopinto: Same as above.

src/web_app/js/util.js
```diff
@@ -74,32 +74,36 @@ function sendUrlRequest(method, url, async, body) {
...
+//       urls: ['turn:1.2.3.4:19305', 'turn:1.2.3.5:19305'],
+//       username: 'username',
+//       credential: 'credential'
+//     },
+//     {
+//       urls: ['stun:stun.example.com:19302']
+//     }
+//   ]
+// }
 function requestTurnServers(turnRequestUrl, turnTransports) {
```
andresusanopinto: Nit: we should rename this to be more about ice servers in general.

andresusanopinto:
lgtm. Though I wonder if the changes on how ice servers are generated breaks any AppRTC android demo.

PR #275, opened 2016-03-04
Provide turn_url from CEOD in params

KaptenJansson:
**Description**
Add back turn_url to params.

**Purpose**
This will allow mobile clients to work until they are updated.


andresusanopinto:
lgtm

PR #276, opened 2016-03-05
Add privacy links

KaptenJansson:
**Description**
screen shot 2016-03-05 at 10 12 08 pm

@samdutton could you PTAL at the CSS in case there is a smarter way of doing it, I'm no CSS expert.
@juberti PTAL.

Not sure about about the position of the links etc.

**Purpose**
Add privacy links

### DIFF
--- a/src/web_app/css/main.css
+++ b/src/web_app/css/main.css
@@ -108,16 +108,15 @@ div#privacy {
   position: absolute;
   bottom: 0;
   right: 0;
-  /* Makes sure to stay behind room links so they can be clicked */
-  z-index: -1
-  margin-right: 1em;
-  margin-bottom: 0.5em;
-  margin-top: calc(100vh - 34em);
+  /* Makes sure to stay behind room links so they can be clicked when the
+      page is small */
+  z-index: -1;
 }
 
 div#privacy a {
+  margin-right: 0.5em;
+  margin-bottom: 0.5em;
   color: gray;
-  margin: 0.5em;
 }
 
 html {
--- a/src/web_app/css/main.css
+++ b/src/web_app/css/main.css
@@ -106,15 +106,14 @@ footer > div.active {
 
 div#privacy {
   position: absolute;
+  color: gray;
   bottom: 7px;
   right: 10px;
-  /* Makes sure to stay behind room links so they can be clicked when the
-      page is small */
-  z-index: -1;
 }
 
 div#privacy a {
   color: gray;
+  font-size: 14px;
 }
 
 html {
--- a/src/web_app/css/main.css
+++ b/src/web_app/css/main.css
@@ -104,6 +104,22 @@ footer > div.active {
   opacity: .8;
 }
 
+div#privacy {
+  position: absolute;
+  bottom: 0;
+  right: 0;
+  /* Makes sure to stay behind room links so they can be clicked */
+  z-index: -1
+  margin-right: 1em;
+  margin-bottom: 0.5em;
+  margin-top: calc(100vh - 34em);
+}
+
+div#privacy a {
+  color: gray;
+  margin: 0.5em;
+}
+
 html {
   height: 100%;
   margin: 0;
--- a/src/web_app/css/main.css
+++ b/src/web_app/css/main.css
@@ -106,16 +106,14 @@ footer > div.active {
 
 div#privacy {
   position: absolute;
-  bottom: 0;
-  right: 0;
+  bottom: 7px;
+  right: 10px;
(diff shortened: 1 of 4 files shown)


src/web_app/html/index_template.html
```diff
@@ -110,6 +110,10 @@
     </svg>
 
   </div>
+  <div id="privacy">
+    <span><a href="//www.google.com/accounts/TOS">Terms</a></span>
```
samdutton: You don't really need to put these links in span elements.

samdutton: I think div#privacy should go in <footer>.

src/web_app/css/main.css
```diff
@@ -104,6 +104,22 @@ footer > div.active {
   opacity: .8;
 }
 
+div#privacy {
+  position: absolute;
+  bottom: 0;
+  right: 0;
+  /* Makes sure to stay behind room links so they can be clicked */
+  z-index: -1
```
samdutton: Might be able to get rid of the following footer > div rule instead (but check it doesn't break div#info-div, etc.):

  width: calc(100% - 20px); 

...then add float:left to the div#sharing-div rules, and add the following:

div#privacy {
  display: block;
  float: right;
  opacity: 1;
}

footer {
    position: absolute;
    bottom: 0;
    width: calc(100% - 20px);
    background: black;
    padding: 10px;
    opacity: .8;
}

(...or something like that!)

src/web_app/css/main.css
```diff
@@ -104,6 +104,22 @@ footer > div.active {
...
+  margin-right: 1em;
+  margin-bottom: 0.5em;
+  margin-top: calc(100vh - 34em);
+}
+
+div#privacy a {
```
samdutton: I'd leave these links blue, like other links, and not use margin (just add a space a pipe between them).

KaptenJansson: @juberti wanted them gray or white.

src/web_app/html/index_template.html
KaptenJansson: The problem is that footer element is only shown when joining the room, we only want to display the links on the room joining page.

KaptenJansson: Will remove span.

samdutton: Ah — forgot that :/.

src/web_app/css/main.css
KaptenJansson: I do not think this is the way to go since we cannot add it to the existing footer as per the comment below, do you have other suggestions for improvement?

samdutton: Fair enough :). 

As is should be OK except I'm wondering what the margins are for. I don't think you'll need these if you set the absolute position with something like `bottom: 30px; right: 20px;`.

KaptenJansson: RIght, I was playing around with those before I used absolute positioning and did not realize I did not need them anymore. They are now removed.

KaptenJansson:
Now it looks like this: screen shot 2016-03-07 at 10 05 49 am

KaptenJansson:
Maybe needs some more margin at the bottom?

samdutton:
You should be able to move away from the bottom right by adding bottom:
30px; right: 20px; (say). But maybe that's not working for you?

KaptenJansson:
ow right you wrote that, will try that ;)

KaptenJansson:
screen shot 2016-03-07 at 10 22 44 am

samdutton:
Hmm... What's the element that contains div#privacy? Maybe you need to add
position: relative; to that.

KaptenJansson:
It's in the body. Why do I need to add relative positioning to the parent element? Does something look out of whack in the screenshot ? ;)

samdutton:
Ignore me — it's all fine :). (I just thought initially you'd added the screenshot to show that there was a problem.)

LGTM.

KaptenJansson:
Aha ;). I thought I would add the screenshot between each change so that people can more easily see the visual result of the change.

juberti:
Terms and Privacy seem too close to one another. Maybe use a "|" in between?

Can you push to a test instance?

KaptenJansson:
Done, https://9-dot-apprtc.appspot.com/

juberti:
I think we want the privacy/terms text to be a little smaller, perhaps 14px instead of 16px.

KaptenJansson:
https://9-dot-apprtc.appspot.com/ updated with the new font size.

KaptenJansson:
Is this okay to commit now?

KaptenJansson:
Merging

PR #287, opened 2016-04-11
Update AppRTC extension to use the network traversal server

KaptenJansson:
**Description**

**Purpose**


andresusanopinto:
Lgtm, but can you write a better cl description?

Also how does the api key restriction by url behaves when doing a request from a chrome app?

KaptenJansson:
Chrome extensions use their ID as an URL for origin/referer.

PR #294, opened 2016-04-28
Typo bug fix

reker123:
Fix the crash when Constants#ICE_SERVER_BASE_URL and the 'ts' param in
request are both empty.

ice_server should be ice_server_base_url, I think it's a typo.


reker123:
I signed it!

reker123:
I signed it!

reker123:
I signed it

PR #295, opened 2016-04-28
Bug fix

reker123:
Fix the crash when Constants#ICE_SERVER_BASE_URL and the 'ts' param in
HTTP Request are both empty.


KaptenJansson:
Thanks for spotting and fixing this! LGTM

PR #297, opened 2016-05-02
Update getStats format

KaptenJansson:
**Description**
Use the modern getStats (shimmed by adapter.js) format + fix error on Firefox.

Most media stats do not work in Firefox. I'd prefer to wait for adapter.js to shim the report name differences before changing anything. I started looking into for AppRTC only but quickly realized it's a deep rabbit hole......

**Purpose**


samdutton:
I'm not knowledgable about getStats, but works for me in stable Chrome and Firefox and code LGTM.

PR #298, opened 2016-05-02
fix vsibr param

### LINKED ISSUE #289: Bitrate parameters (vsibr, vsbr) don't work.
These parameters result in the SDP being munged, adding fmtp parameters for the VP8 codec. But if the codec used isn't VP8 (it's currently VP9 by default), this doesn't do anything.

KaptenJansson:
**Description**
The codec in use was not considered when applying the `vsibr` param. This change will apply the bitrate to whatever video codec is used.

**Purpose**
Fixes #289


samdutton:
LGTM.

PR #311, opened 2016-05-30
Use srcObject instead of src

KaptenJansson:
**Description**
Chrome now has srcObject natively in 52 which means the adapter shim is no longer used. Since AppRTC was setting src to '' rather than srcObject = null it stopped working.

**Purpose**
Fix rejoin/new room bug: https://bugs.chromium.org/p/chromium/issues/detail?id=614181


KaptenJansson:
Merging now, can rubber stamp this later ;).

samdutton:
Rubber Stamp :).

PR #315, opened 2016-06-13
Only remove FEC if videofec=false.

stefanholmer:
**Description**

**Purpose**

### DIFF
--- a/src/web_app/js/sdputils.js
+++ b/src/web_app/js/sdputils.js
@@ -248,8 +248,9 @@ function removeCodecByPayloadType(sdpLines, payloadType) {
 }
 
 function maybeRemoveVideoFec(sdp, params) {
-  if (params.videoFec === 'true')
+  if (params.videoFec === 'true') {
     return;
+  }
 
   var sdpLines = sdp.split('\r\n');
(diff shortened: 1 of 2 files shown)


src/web_app/js/sdputils.js
```diff
@@ -247,7 +247,10 @@ function removeCodecByPayloadType(sdpLines, payloadType) {
   return sdpLines;
 }
 
-function maybeRemoveVideoFec(sdp) {
+function maybeRemoveVideoFec(sdp, params) {
+  if (params.videoFec === 'true')
```
KaptenJansson: please use curly braces (linter will complain ;))

stefanholmer:
Not sure why travis complains, it doesn't complain locally?

KaptenJansson:
Seems like we have managed to check in a lint error, https://github.com/stefanholmer/apprtc/blob/cbd5fe6c75bea91783c88f4406f059c6f5b500cc/src/web_app/js/sdputils.js#L51 These should be indented like the switch case above.

KaptenJansson:
I will fix the lint errors separately.

PR #333, opened 2016-07-25
Fix bind bindMST function, use ontrack and replace attach/replaceMediaStream with srcObject

### LINKED ISSUE #331: bindMstToUserIdForCallstats_ should not be called onaddstream
**Description**

The code added in the commit #261 included that function and broke webkit client.

onaddstream is called "nearly immediately after the remote end of the connection is set using RTCPeerConnection.setRemoteDescription()":

https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/onaddstream

But the method is using some peer connection values such as the localDescription that it could be not initialized.

Probably a solution is calling the same function in the promise of createAnswer? That way all the information should be there.

Also using a deprecated API for new code is not a good idea.

KaptenJansson:
**Description**
Fixes #331 

**Purpose**

### DIFF
--- a/src/web_app/js/appcontroller.js
+++ b/src/web_app/js/appcontroller.js
@@ -8,7 +8,7 @@
 
 /* More information about these options at jshint.com/docs/options */
 
-/* globals adapter, trace, InfoBox, setUpFullScreen, isFullScreen,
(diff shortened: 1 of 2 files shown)


KaptenJansson:
Merging this in order to the get latest callstats pushed.

PR #302, opened 2016-05-16
Add support for GET parameter to disable video FEC.

stefanholmer:
**Description**
Adds a GET parameter to enable filtering out the FEC/RED codecs so that apprtc can be used without FEC.

**Purpose**
Mostly for testing, but can also be useful in some situations for users who don't want to waste bandwidth on FEC.

PR #343, opened 2016-08-02
Fixing issues with option to remove video FEC.

taylor-b:
1. Only remove if the videofec GET parameter is "false". Don't remove
   by default (if there's no GET parameter).
2. If it *is* removed, remove it from both the local and remote
   descriptions. Otherwise the remote PeerConnection thinks FEC is
   negotiated, but the local PeerConnection doesn't and is surprised
   when it sees a FEC packet.


KaptenJansson:
Thanks for fixing, when I reviewed the FEC changes in #302 I did not think of this. We should probably have a test for this. Will create an issue.

KaptenJansson:
LGTM

Issue #312, opened 2016-06-02
Start using DTLS for loopback calls


KaptenJansson:
Since SDES should not be used and we are hitting Chrome [issues](https://bugs.chromium.org/p/chromium/issues/detail?id=616263#c9) we should update AppRTC to use DTLS instead. This means we have to create two peerConnections since when using DTLS one peerConnection cannot act as a client and server.

I'm open for suggestions on how to solve this but I was thinking that we could create and manage the 2nd peerConnection in [loopback.js](https://github.com/webrtc/apprtc/blob/fixLoopback/src/web_app/js/loopback.js). I guess it's also possible to manage it in call.js but there is already a lot going in there.

@juberti @samdutton @jiayliu

KaptenJansson:
This needs to revisited.

fippo:
ping. SDES is going away...

PR #346, opened 2016-08-05
Add real loopback calling, improve callstats initialization and some style fixes.

### LINKED ISSUE #34: AppRTC loopback should work with DTLS
Quoth https://github.com/GoogleChrome/webrtc/blob/master/samples/web/content/apprtc/apprtc.py#L395

debug = self.request.get('debug')
if debug == 'loopback':
  # Set dtls to false as DTLS does not work for loopback.
  dtls = 'false'
Indeed, failing to disable DTLS causes errors like:
Error(transport.cc:101): Answerer must use either active or passive value for setup attribute.

Since SDES is eventually going to disappear in favor of DTLS we should make sure loopback calls work with DTLS as well. The answer is probably to set up a second PeerConnection on the same page.
Once the JS version is updated the same should be done for the Android & iOS sample apps.

KaptenJansson:
**Description**
* Fixes #34 #312 (Removes single peerconnection loopback using SDES which is deprecated)
* Setup a 2nd Call constructor without any media to act as a loopback peerconnection which clones the remote stream and adds it back as a local stream which the 1st peerconnection receives and processes as usual.
* Make the callstats init more robust (I know it's a bit messy to have it in the same PR but I encountered lots of issues while testing this thus had to fix it).
* Also fixed the onstream test to use ontrack instead.
* Fixed the transition for rejoin and new room calls (it never flipped back the video tags between calls)
* Fixed some indentations here and there.

**Purpose**
Deprecate SDES loopback and make callstats init more robust.

### DIFF
--- a/src/app_engine/apprtc.py
+++ b/src/app_engine/apprtc.py
@@ -256,7 +256,6 @@ def get_room_parameters(request, room_id, client_id, is_initiator):
 
   debug = request.get('debug')
   if debug == 'loopback':
-    # Set dtls to false as DTLS does not work for loopback.
     include_loopback_js = '<script src="/js/loopback.js"></script>'
   else:
     include_loopback_js = ''
@@ -529,7 +528,6 @@ class JoinPage(webapp2.RequestHandler):
 
   def post(self, room_id):
     is_loopback = self.request.get('debug') == 'loopback'
-
     if is_loopback:
       client_id = constants.LOOPBACK_CLIENT_ID
       memcache_client = memcache.Client()
--- a/src/app_engine/apprtc.py
+++ b/src/app_engine/apprtc.py
@@ -257,7 +257,6 @@ def get_room_parameters(request, room_id, client_id, is_initiator):
   debug = request.get('debug')
   if debug == 'loopback':
     # Set dtls to false as DTLS does not work for loopback.
-    dtls = 'false'
     include_loopback_js = '<script src="/js/loopback.js"></script>'
   else:
     include_loopback_js = ''
@@ -388,8 +387,6 @@ def add_client_to_room(request, room_id, client_id, is_loopback):
     if occupancy == 0:
       is_initiator = True
       room.add_client(client_id, Client(is_initiator))
-      if is_loopback:
-        room.add_client(constants.LOOPBACK_CLIENT_ID, Client(False))
     else:
       is_initiator = False
       other_client = room.get_other_client(client_id)
@@ -410,7 +407,7 @@ def add_client_to_room(request, room_id, client_id, is_loopback):
     else:
       retries = retries + 1
   return {'error': error, 'is_initiator': is_initiator,
-          'messages': messages, 'room_state': str(room)}
+          'messages': messages, 'room_state': str(room), 'client_id': client_id}
 
 def remove_client_from_room(host, room_id, client_id):
   key = get_memcache_key_for_room(host, room_id)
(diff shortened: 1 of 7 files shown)


KaptenJansson:
@hkjellander could you PTAL?

src/app_engine/apprtc.py
```diff
@@ -256,8 +256,6 @@ def get_room_parameters(request, room_id, client_id, is_initiator):
 
   debug = request.get('debug')
   if debug == 'loopback':
-    # Set dtls to false as DTLS does not work for loopback.
-    dtls = 'false'
```
hkjellander: Please include something in the PR description about this so one doesn't have to look up #312 and #34

src/web_app/js/call.js
```diff
@@ -16,7 +16,7 @@
 
 'use strict';
 
-var Call = function(params) {
+var Call = function(params, doNotRequestMediaAndIceServers) {
```
hkjellander: Is this called somewhere else than in https://github.com/webrtc/apprtc/blob/6d86424372f85b2e2295126e9ebbf0ddc7f6348c/src/web_app/js/loopback.js#L27 ? Won't that break then?

src/app_engine/apprtc.py
KaptenJansson: Done

src/web_app/js/call.js
KaptenJansson: This is called in [appcontroller.js](https://github.com/webrtc/apprtc/blob/master/src/web_app/js/appcontroller.js#L148) but if it's not provided when creating the constructor, media and ice servers will be requested which was the default behavior before hence all places where it's called like before (just with params) the default behavior will be preserved.

src/web_app/js/call.js
```diff
@@ -492,6 +491,13 @@ Call.prototype.startSignaling_ = function() {
     } else {
       this.pcClient_.startAsCallee(this.params_.messages);
     }
+    // Setup the remote loopback peerconnection client. Give the first
+    // peerConnection some time to setup.
+    if (this.params_.isLoopback && this.params_.isInitiator) {
+      setTimeout(function() {
+        setupLoopback(this.params_);
+      }.bind(this), 2000);
```
hkjellander: "some time" is vague. How do we know 2000 ms is enough for whatever we're waiting for?

src/web_app/js/loopback.js
```diff
@@ -12,74 +12,65 @@
...
+  // peerConnection 1 has added it.
+  call.onremotestreamadded = function(event) {
+    if (this.params_.clientId === 'LOOPBACK_CLIENT_ID_2') {
+      // Since this fired once per track we need to make sure it will only add
+      // the stream once.
+      if (this.pcClient_.loopBackStream === null) {
+        // 2nd peerConnection should not be the initiator.
+        if (!this.params_.isInitiator_) {
+          this.pcClient_.loopBackStream = event.clone();
+          // Disable tracks on the remote tracks of the remote stream otherwise
```
hkjellander: Change to "Disable audio tracks on the remote stream otherwise..." ?

hkjellander:
lgtm.
Next time; please split up such changes to multiple PRs instead.

src/web_app/js/call.js
KaptenJansson: Turns out this was not needed anymore, I guess I had race conditions when I started but they have now since been eliminated.

src/web_app/js/loopback.js
KaptenJansson: Done

KaptenJansson:
Yeah will split it up next time! Thanks for the review.

PR #358, opened 2016-09-05
Adding back callstats improvements from PR#346

KaptenJansson:
**Description**
Checks if the key and ID are present properly (did not cover all cases previously).

**Purpose**
Ensure that we do not try to use callstats API's when not initialized.


KaptenJansson:
Merging this as part of reverting the new loopback logic in #346 which included these tweaks which also got reverted.

PR #364, opened 2016-09-09
Escape HTML in error message and update room server URL

KaptenJansson:
**Description**
Escape HTML in error message
Update room server URL

**Purpose**
Fix XSS and update room server URL


KaptenJansson:
@alvestrand could you PTAL?

alvestrand:
Even though I've never written go, LGTM.

PR #371, opened 2016-09-16
Replace jstest with Karma and Jasmine

KaptenJansson:
**Description**

- Replace [jstestdriver + phantomjs](https://www.npmjs.com/package/grunt-jstestdriver) with [Karma](https://karma-runner.github.io/1.0/index.html) and [Jasmine](http://jasmine.github.io/).
- Replace jscs and jshint with eslint (the same we use in the other repos). Made sense since I had to rewrite much of the code anyway.
- Configure travis to run the unit tests on Chrome stable, beta, dev and Firefox stable. Good thing about running on Firefox is that it helps catching deprecated API's since it's usually more strict about using deprecated API's compared to Chrome. Can disable FF and some of the Chrome channels if the tests take to long.

**Purpose**
Run the unit tests on a modern browser instead of a phantomjs version from 2011. The jstestdriver package has not updated the phantomjs version in a very long time.


KaptenJansson:
Could you PTAL? I know it's a big PR but not much choice here.

KaptenJansson:
Also, there is a new Github review UI, you select the type of review in the right corner after clicking on the files changed tab in the PR.

build/get_python_test_deps.py
```diff
@@ -76,10 +102,24 @@ def DownloadWebTestIfNecessary():
   _Download(WEBTEST_URL, webtest_file)
   _Untar(webtest_file)
 
+# Downloads the requests python module used to download other deps.
+def DownloadRequestsIfNecessary():
+  request_file = 'requestsPythonModule'
```
hkjellander: Can't we use a more standardized way to download this instead of creating our own way? Can't bower do it for us?

src/web_app/js/appcontroller.js
```diff
@@ -531,9 +531,9 @@ AppController.prototype.setIconTimeout_ = function() {
     window.clearTimeout.bind(this, this.hideIconsAfterTimeout);
   }
   this.hideIconsAfterTimeout =
-    window.setTimeout(function() {
-    this.hideIcons_();
-  }.bind(this), 5000);
+      window.setTimeout(function() {
```
hkjellander: Restore indentation?

src/web_app/js/appcontroller_test.js
```diff
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
+ *  Copyright (c) 2016 The WebRTC project authors. All Rights Reserved.
```
hkjellander: Year should only be 2016 on new files added this year.

build/.eslintrc
```diff
@@ -0,0 +1,63 @@
+{
```
KaptenJansson: Bower is not used in this project, rather npm. Yes it does match the Google style guide, taken from the samples [repo](https://github.com/webrtc/samples/blob/gh-pages/.eslintrc).

build/get_python_test_deps.py
```diff
@@ -53,8 +60,27 @@ def _Untar(path):
   try:
     tar_file.extractall()
   finally:
+    dir_name = tar_file.getnames()
     tar_file.close()
-
+      # Return the first member in the returned list as it's the top dir
```
KaptenJansson: Will fix.

build/get_python_test_deps.py
```diff
@@ -53,8 +60,27 @@ def _Untar(path):
...
+      # functions.
+    return dir_name[0]
+
+# Installs the requests python module.
+def _InstallRequests(requests_dir):
+  cwd = os.getcwd()
+  try:
+    print 'About to install requests module into your system python.'
+    os.chdir(requests_dir)
+    result = os.system('sudo python setup.py install')
```
KaptenJansson: Yes there is but the reason I did it this way is because existing code already does it this way but will look into it.

build/get_python_test_deps.py
KaptenJansson: pip is the standard way of download and installing python modules but I thought not everyone had pip but I know realize that it might be standard on most linux dists (and Mac it seems). Will look into this.

src/web_app/js/appcontroller_test.js
KaptenJansson: So we do not update the year on modified files?

hkjellander: Right, that's generally the practice: https://chromium.googlesource.com/chromium/src/+/master/styleguide/c++/c++.md#File-headers
"$YEAR should be set to the current year at the time a file is created, and not changed thereafter."

It makes life easier for everyone, especially after new year's eve.

KaptenJansson: aha I see. Will fix.

KaptenJansson:
@hkjellander PTAL again now. I've now finally managed to get it all working again.

build/get_python_test_deps.py
KaptenJansson: Now downloading webtest and requests using pip from within python. Also it's installed to a user accessible location (no sudo required)

build/get_python_test_deps.py
KaptenJansson: Now using pip as mentioned in the above comment.

.gitignore
```diff
@@ -1,11 +1,12 @@
 *.pyc
 .DS_Store
+browser*
```
KaptenJansson: Done

build/get_python_test_deps.py
```diff
@@ -11,7 +11,14 @@
...
 GAE_UPDATECHECK_URL = 'https://appengine.google.com/api/updatecheck'
 WEBTEST_URL = 'https://nodeload.github.com/Pylons/webtest/tar.gz/master'
+REQUESTS_URL = 'https://github.com/kennethreitz/requests/tarball/master'
+TEMP_DIR = 'temp'
 
+def setUpTempDir():
+  if not os.path.exists(TEMP_DIR):
+    os.mkdir(TEMP_DIR)
+
+  os.chdir(TEMP_DIR)
```
KaptenJansson: Removed.

src/web_app/js/appcontroller.js
```diff
@@ -530,8 +530,7 @@ AppController.prototype.setIconTimeout_ = function() {
   if (this.hideIconsAfterTimeout) {
     window.clearTimeout.bind(this, this.hideIconsAfterTimeout);
   }
-  this.hideIconsAfterTimeout =
-    window.setTimeout(function() {
+  this.hideIconsAfterTimeout = window.setTimeout(function() {
     this.hideIcons_();
   }.bind(this), 5000);
```
hkjellander: What guarantee do we have that 5000 ms is enough?

src/web_app/js/call_test.js
```diff
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
+ *  Copyright (c) 2016 The WebRTC project authors. All Rights Reserved.
```
hkjellander: Don't change year.

hkjellander:
And in the future, I will reject such a large PR - please split up and do changes like this in smaller steps (some of the changes were purely formatting/lint changing).

src/web_app/js/appcontroller.js
KaptenJansson: Not sure, I did not change that, I only moved the variable up a line.

KaptenJansson: But this is the time before hiding the buttons in the apprtc call UI.

KaptenJansson:
Yes will do, there were so many things depending on each other. I did split it up a little in separate commits but the review UI might not handle that in a good way.

package.json
```diff
@@ -9,21 +9,28 @@
...
+    "jasmine-core": "^2.4.1",
+    "karma": "^1.1.0",
+    "karma-chrome-launcher": "^1.0.1",
+    "karma-cli": "^1.0.1",
+    "karma-eslint": "^2.1.0",
+    "karma-firefox-launcher": "^1.0.0",
+    "karma-jasmine": "^1.0.2",
     "socket.io-client": "^1.2.0",
-    "webrtc-adapter": "^1.0.6"
+    "webrtc-adapter": "^2.0.0",
```
KaptenJansson: Good catch

build/start-tests.sh
```diff
@@ -0,0 +1,22 @@
+#
+#  Copyright (c) 2016 The WebRTC project authors. All Rights Reserved.
+#
+#  Use of this source code is governed by a BSD-style license
+#  that can be found in the LICENSE file in the root of the source
+#  tree.
+#!/bin/sh
```
KaptenJansson: Done

Gruntfile.js
```diff
@@ -37,34 +37,20 @@ module.exports = function(grunt) {
...
         command: 'python build/get_python_test_deps.py'
       },
-      installPythonTestDepsOnLinux: {
-        command: 'python build/install_webtest_on_linux.py webtest-master/'
-      },
       runPythonTests: {
-        command: ['python', 'build/run_python_tests.py', 'google_appengine/',
-                  out_app_engine_dir, 'webtest-master/'].join(' ')
+        command: ['python', 'build/run_python_tests.py',
+                  'temp/google_appengine/', out_app_engine_dir].join(' ')
```
KaptenJansson: I added temp, tmp was most likely used in the existing code. Will unify.

build/get_python_test_deps.py
```diff
@@ -38,48 +36,43 @@ def _Download(url, to):
...
-  tar_file = tarfile.open(path, 'r:gz')
+def Install(package):
   try:
-    tar_file.extractall()
-  finally:
-    tar_file.close()
+    print 'Installing python package using pip: ' + package
+    pip.main(['install', '--user' , package])
+  except OSError:
+    print 'Could not install %s due to : %s' % package % OSError
```
KaptenJansson: aha sorry, not sure why the linter did not catch this. And shame on me for not testing it properly.

Gruntfile.js
KaptenJansson: This is the same temp dir as in build/get_python_test_deps.py (which will be a const in that file). Not sure if a constant makes sense here since they are in different files and different languages.

PR #372, opened 2016-09-16
Redirect if URL is not appr.tc and redirect is not handled by ISP

### LINKED ISSUE #221: www.appr.tc, apprtc.net, should redirect to appr.tc

KaptenJansson:
**Description**
Redirect old URL's not handled by the ISP to appr.tc and make sure the path and URL paramters tag along.

Fixes #221

**Purpose**
Make everyone use appr.tc.

### DIFF
--- a/src/app_engine/apprtc.py
+++ b/src/app_engine/apprtc.py
@@ -553,8 +553,7 @@ class MainPage(webapp2.RequestHandler):
 
   def get(self):
     """Renders index.html."""
-    if self.request.headers['Host'] == 'apprtc.net':
-      webapp2.redirect('https://www.apprtc.net', permanent=True)
+    checkIfRedirect(self);
     # Parse out parameters from request.
     params = get_room_parameters(self.request, None, None, None)
     # room_id/room_link will not be included in the returned parameters
@@ -569,6 +568,7 @@ class RoomPage(webapp2.RequestHandler):
 
   def get(self, room_id):
     """Renders index.html or full.html."""
+    checkIfRedirect(self)
     # Check if room is full.
     room = memcache.get(
         get_memcache_key_for_room(self.request.host_url, room_id))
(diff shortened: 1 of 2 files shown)


samdutton:
I'm not a Python boffin, but LGTM.

src/app_engine/apprtc.py
```diff
@@ -590,6 +590,20 @@ def get(self):
     params = get_room_parameters(self.request, None, None, None)
     self.response.write(json.dumps(params))
 
+def checkIfRedirect(self):
+  REDIRECT_DOMAINS = [
```
hkjellander: Would it make sense to move this into app_engine/constants.py even if it's only used here. Buried 600 lines down in source code is a pretty hidden place.

src/app_engine/apprtc.py
```diff
@@ -590,6 +590,20 @@ def get(self):
...
+    'apprtc.appspot.com', 'apprtc.webrtc.org', 'www.appr.tc'
+  ]
+  ARGUMENTS = self.request.arguments()
+  parsed_args = ''
+  if self.request.headers['Host'] in REDIRECT_DOMAINS:
+    webapp2.redirect('https://appr.tc', permanent=True)
```
hkjellander: Why do a redirect here and then another one at line 606? It seems the one at line 606 is the correct one and that this one should be removed. Or have a I misunderstood how webapp2.redirect works? I tried finding documentation but couldn't find anything pointing at any reason using it like this.

src/app_engine/apprtc.py
```diff
@@ -590,6 +590,20 @@ def get(self):
     params = get_room_parameters(self.request, None, None, None)
     self.response.write(json.dumps(params))
 
+def checkIfRedirect(self):
+  REDIRECT_DOMAINS = [
+    'apprtc.appspot.com', 'apprtc.webrtc.org', 'www.appr.tc'
+  ]
+  ARGUMENTS = self.request.arguments()
```
hkjellander: The ARGUMENTS constant doesn't add anything useful, just use self.request.arguments() directly at line 601 instead.

src/app_engine/apprtc.py
```diff
@@ -590,6 +590,20 @@ def get(self):
...
+  parsed_args = ''
+  if self.request.headers['Host'] in REDIRECT_DOMAINS:
+    webapp2.redirect('https://appr.tc', permanent=True)
+    for argument in ARGUMENTS:
+      if parsed_args == '':
+        parsed_args += '?'
+      parsed_args += argument + '&'
+    redirect_url = 'https://appr.tc' + self.request.path + parsed_args
```
hkjellander: Move 'https://appr.tc' into a constant in app_engine/constants.py

src/app_engine/apprtc.py
```diff
@@ -590,6 +590,20 @@ def get(self):
     params = get_room_parameters(self.request, None, None, None)
     self.response.write(json.dumps(params))
 
+def checkIfRedirect(self):
+  REDIRECT_DOMAINS = [
+    'apprtc.appspot.com', 'apprtc.webrtc.org', 'www.appr.tc'
```
hkjellander: Out of curiosity I also tried www.apprtc.appspot.com and it gives me an "invalid certificate" error. Is that something we want to fix?

src/app_engine/apprtc.py
KaptenJansson: Done

src/app_engine/apprtc.py
KaptenJansson: Done

src/app_engine/apprtc.py
KaptenJansson: I guess we forgot to include www. in the certificate. Not sure if this is important enough to warrant us to create a new certificate. @juberti @samdutton  WDYT?

Issue #385, opened 2016-10-14
AppRTC no longer allows mic-only and cam-only Firefox users to participate (regression)


jan-ivar:
**Browsers and versions affected**
Firefox 49 (release) and newer

**Description**
Starting with Firefox 49, a user with only a mic and no camera can no longer talk on an AppRTC call (presumably the reverse is true as well, a user with camera but no mic can't be seen).

Firefox 49 updated to follow [the spec](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediadevices-getusermedia), which says to fail an audio+video request unless both are available.

**Steps to reproduce**
Join an appRTC call from a machine with mic only and no camera.

**Expected results**
- Joins conversation by sending audio only, and still receives audio+video whatever.

**Actual results**
- "Failed to get access to local media. Error name was NotFoundError. Continuing without sending a stream." [OK]
- Joins conversation without sending audio (still receives audio+video whatever).

KaptenJansson:
I've pushed a new release of AppRTC with the latest adapter release, could you please verify?

jan-ivar:
WFM, Thanks!

KaptenJansson:
No problem and thanks for the fix!

PR #386, opened 2016-10-14
Allow mic-only & cam-only users in spec-compliant browsers - broke FF49

jan-ivar:
**Description**
Fix for https://github.com/webrtc/apprtc/issues/385.

**Purpose**
Modifies getUserMedia call to allow mic-only and cam-only users (in Firefox 49+)


KaptenJansson:
LGTM!

Thanks for doing this, I've been meaning to do it for a while but it got lost.

PR #389, opened 2016-10-26
Add support fordatachannel

hekra01:
**Description**
Add support for datachannel.
Related to
* https://github.com/njovy/AppRTCDemo/pull/32
* https://bugs.chromium.org/p/webrtc/issues/detail?id=6647

**Purpose**
Will enable to send text message for example.

### DIFF
--- a/src/web_app/js/call.js
+++ b/src/web_app/js/call.js
@@ -587,3 +587,7 @@ Call.prototype.onError_ = function(message) {
     this.onerror(message);
   }
 };
+
+Call.prototype.sendData = function(message) {
+  this.pcClient_.sendData(message);
+};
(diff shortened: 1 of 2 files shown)


KaptenJansson:
@samdutton @juberti what's your opinion on adding data channels to AppRTC? Note that this does not include an UI or anything so more work would be needed for this to be usable by normal users.

KaptenJansson:
@samdutton @juberti ping

samdutton:
For chat?

On 2 Dec 2016 9:36 a.m., "Christoffer Jansson" <notifications@github.com>
wrote:

@samdutton <https://github.com/samdutton> @juberti
<https://github.com/juberti> ping

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/webrtc/apprtc/pull/389#issuecomment-264412355>, or mute
the thread

hekra01:
Yes, chat is one use case. Could also be used for file transfer etc...

KaptenJansson:
I think then dataChannel should only be setup when one of the features are enabled/requested however that will require renegotiation off course which AppRTC does not currently handle.

KaptenJansson:
Closing. A followup PR with datachannel and a feature that uses it (e.g chat with a UI etc) will be considered.

PR #390, opened 2016-11-02
Adding back TLS1.x ciphers to increase compatability

KaptenJansson:
**Description**
Some WebSocket libraries when used on older Android versions do not support TLS 1.2.

**Purpose**
Increase comparability....


hkjellander:
I guess I'll wait for you to fix the Travis issue?

KaptenJansson:
Interesting, I do not see that error locally. Also, this is totally unrelated to my changes, eslint has probably been updated.

PR #393, opened 2016-11-06
handle room full state by returning to room selection state.

hekra01:
**Description**
Handle room full state by returning to room selection state.
Once a room is full, the same user interface will be displayed, whether  doing POST `roomserver/join/roomId` or GET `roomserver/r/roomId`

**Purpose**
When a room is full, new requests to join the room result in the same user interface


KaptenJansson:
Thanks for doing this, LGTM. (will merge soonish)

PR #406, opened 2016-12-23
Check that the candidateObject is an object and has a candidate property

KaptenJansson:
**Description**

**Purpose**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ browser*
 firefox*
 node_modules
 out
+rtstats.js
 secrets.json
 temp/*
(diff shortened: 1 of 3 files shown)


MirkoBonadei:
lgtm on "src/web_app/js/peerconnectionclient.js".
We are going to do another pull request for the other fix.

fippo:
that is mostly used for ?it=relay, right? If so lets drop the code in favor of setting iceTransport

KaptenJansson:
I will create a more complete PR later, closing this for now.

PR #407, opened 2016-12-23
Trying to use SSL with SNI ext on older python 2.x

MirkoBonadei:
**Description**
Ensuring certificate verification for old python 2.x versions.

**Purpose**

@KaptenJansson this is a little bit hacky because we are installing modules which will then be used by the script itself (see the global imports). If you think that a refactoring of this script is useful please let me know and I will try to split it into 2 different files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ matrix:
 
 before_install:
   - npm install -g npm@'>=3.0.0'
+  - sudo apt-get install -y build-essential libssl-dev libffi-dev python-dev
 
 before_script:
   - npm install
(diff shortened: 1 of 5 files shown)


MirkoBonadei:
Unfortunately `--verbose` fills up the travis output buffer. I will get back on this issue on Tuesday.

MirkoBonadei:
@KaptenJansson Ok, we are green and with certificate verification enabled.
Unfortunately I had to follow a try and error approach to debug the build on TravisCI so I will squash this branch in a unique commit before merging.

PR #425, opened 2017-03-01
Improved app.yaml

mahiuchun:
Latest documentation on app.yaml recommends removing application and version element.
https://cloud.google.com/appengine/docs/standard/python/config/appref
At least for my case, gcloud would complain if these elements are present.

For setting up environment variables, I haven't found gcloud equivalent of appcfg.py --env_variable.
Since it is not clear for someone not familiar with app engine that appcfg.py is not available from the new Google Cloud SDK, I would say removing these instructions reduces confusion.

### DIFF
--- a/src/app_engine/app.yaml
+++ b/src/app_engine/app.yaml
@@ -1,5 +1,3 @@
-application: apprtc
-version: 9
 runtime: python27
 threadsafe: true
 api_version: 1


src/app_engine/app.yaml
```diff
@@ -55,10 +53,6 @@ libraries:
 env_variables:
   BYPASS_JOIN_CONFIRMATION: false
   # Only change these while developing, do not commit to source!
-  # Use appcfg.py --env_variable=CALLSTATS_APP_ID:ID  \
```
KaptenJansson: We want to keep these as we do use them when we deploy AppRTC. This will save us from having internal documentation. But you are correct that gcloud does not have a way to do this hence why we still use appcfg to deploy. Gcloud is used everywhere else though.

KaptenJansson (changes requested):

mahiuchun:
Done

KaptenJansson (approved):

KaptenJansson:
LGTM

Issue #437, opened 2017-04-28
wrong code make apprtc not work without ice server


danhdevelop:
The code in apprtc.py at line 282:

```
if len(ice_server_base_url) > 0:
    ice_server_url = constants.ICE_SERVER_URL_TEMPLATE % \
        (ice_server_base_url, constants.ICE_SERVER_API_KEY)
else:
    ice_server_base_url = ''
```
in the else should be: `ice_server_url=''`

KaptenJansson:
Thanks for filing, this is now fixed.

PR #438, opened 2017-04-28
Fixing variable name

MirkoBonadei:
**Description**
As reported in https://github.com/webrtc/apprtc/issues/437, AppRTC does not work without an ICE server because
of a wrong assignment.

This pull reuqest should fix.

**Purpose**
Fixing https://github.com/webrtc/apprtc/issues/437.


KaptenJansson:
LGTM

PR #448, opened 2017-05-31
Unused files: enums.js and adapter.js?

itsazzad:
**Description**
The following files not exist:
```
            'src/web_app/js/enums.js',
            'src/web_app/js/adapter.js',
```

**Purpose**
Remove unused files entires.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -18,15 +18,13 @@ module.exports = function(grunt) {
         options: {
           import: 2
         },
-        src: ['src/**/*.css'
-        ]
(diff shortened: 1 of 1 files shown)


KaptenJansson:
https://github.com/webrtc/apprtc/blob/master/build/gen_js_enums.py generates enum.js but you are correct in that 'src/web_app/js/adapter.js', no longer exists since we now use 'node_modules/webrtc-adapter/out/adapter.js' instead.

KaptenJansson:
Ping

PR #449, opened 2017-06-02
update node to 6

KaptenJansson:
**Description**

**Purpose**


src/web_app/js/appcontroller_test.js
```diff
@@ -80,10 +85,16 @@ describe('AppControllerTest', function() {
         .toBeFalsy();
   });
 
-  it('Hide UI after clicking the join button', function() {
+  it('Hide UI after clicking the join button', function(done) {
     // Verifies that the UI is hidden after clicking the button.
-    $(UI_CONSTANTS.confirmJoinButton).click();
-    expect($(UI_CONSTANTS.confirmJoinDiv).classList.contains('hidden'))
-        .toBeTruthy();
+    // Need to wait a few seconds for the element to update it's class list.
```
MirkoBonadei: Is this because of the race condition you were telling me the other day?

In that case let's add some more context in the comment. We never know if we will need to look into this again in the future.

KaptenJansson: Yeah I believe the classList for the element is not updated fast enough. Will add more info.

PR #461, opened 2017-06-14
change `TURN_SERVER_OVERRIDE` into `ICE_SERVER_OVERRIDE`, and return it as `pc_config.iceServers`

### LINKED ISSUE #456: How to use TURN_SERVER_OVERRIDE config in AppRTC-Android client?
**Browsers and versions affected**
AppRTC-Android,  https://chromium.googlesource.com/external/webrtc/+/832e67a2c35f20de648bc4b532ef15092ae912f0

**Description**
The [AppRTC-Android client](https://chromium.googlesource.com/external/webrtc/+/832e67a2c35f20de648bc4b532ef15092ae912f0/webrtc/examples/androidapp/src/org/appspot/apprtc/RoomParametersFetcher.java#132) would expects `pc_config.iceServers` field from response for TURN and STUN server urls, but [apprtc server](https://github.com/webrtc/apprtc/blob/master/src/app_engine/apprtc.py#L47) will always return empty `pc_config.iceServers`, should I modify AppRTC-Android code or apprtc server code? Thanks!

**Steps to reproduce**

**Expected results**

**Actual results**

Piasy:
**Description**

Fix #456 

**Purpose**

### DIFF
--- a/src/app_engine/apprtc.py
+++ b/src/app_engine/apprtc.py
@@ -44,10 +44,10 @@ def get_hd_default(user_agent):
 # iceServers will be filled in by the TURN HTTP request.
 def make_pc_config(ice_transports):
   config = {
-  'iceServers': [],
+  'iceServers': constants.ICE_SERVER_OVERRIDE,
   'bundlePolicy': 'max-bundle',
   'rtcpMuxPolicy': 'require'
-  };
+  }
   if ice_transports:
     config['iceTransports'] = ice_transports
   return config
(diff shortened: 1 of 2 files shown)


Piasy:
Actually it doesn't work, I setup another ICE server to return TURN/STUN server info, and with that ICE server, it finally works. I'm going to close it.

KaptenJansson:
I see, will look into to why the override does not work, it did when I added it.

PR #470, opened 2017-07-05
Missed to remove the deps when creating the callstats constructor

KaptenJansson:
**Description**

**Purpose**


KaptenJansson:
No worries I missed it too. Also I should really add linting on pre-submit......

PR #482, opened 2017-09-12
peerconnectionclient: use addTrack instead of addStream

fippo:
uses addTrack instead of addStream which has been removed from the spec.
Using addTrack here would have probably (hopefully) caught
    https://bugs.chromium.org/p/chromium/issues/detail?id=762527

### DIFF
--- a/src/web_app/js/peerconnectionclient.js
+++ b/src/web_app/js/peerconnectionclient.js
@@ -82,7 +82,9 @@ PeerConnectionClient.prototype.addStream = function(stream) {
   if (!this.pc_) {
     return;
   }
-  this.pc_.addStream(stream);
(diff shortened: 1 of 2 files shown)


fippo:
hrm. The test mockup seems slightly broken...

KaptenJansson:
Yeah you need to add addTrack mock instead of https://github.com/fippo/apprtc/blob/7ec041e10fcc7f258ca79c27d08ed79139f97df5/src/web_app/js/peerconnectionclient_test.js#L38

fippo:
now I need to rebase?

KaptenJansson:
I've merged my PR.

KaptenJansson:
I will merge and fix this part of fixing the other tests.

PR #489, opened 2017-10-10
Remove incorrect and unused dep

KaptenJansson:
**Description**
Default python on travis was 2.7.6 and we need 2.7.13 to get SNI support. I had to change language from nodejs to python and then manually install nodejs. This resulted in travis using virtualenv hence I have to check if it's running on travis or not and toggle the --user flag for pip.
**Purpose**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 dist: trusty
-language: node_js
-node_js:
-  - 6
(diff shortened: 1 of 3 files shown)


KaptenJansson:
I will fix the tests in a separate PR since they were broken before this PR.

PR #490, opened 2017-10-12
Revert "Merge pull request #482 from fippo/addtrack"

KaptenJansson:
This reverts commit 5e0928a14c8d1b24010d0f97f1dc4822b612cf2d, reversing
changes made to db6a194507c61d0a9f683d24405749a4f90f9fac.

**Description**
CC @fippo 

Bad review by me...

**Purpose**

PR #491, opened 2017-10-12
peerconnectionclient: (fix) undefined "this" in forEach over getTracks()

VinceOPS:
Hi there!

**Description**
`this` is undefined in the `forEach` loop responsible for filling PeerClient's tracks.
Indeed, as stated on [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach), 

> If a thisArg parameter is provided to forEach(), it will be used as callback's this value.  Otherwise, the value undefined will be used as its this value

**Purpose**
Avoid the **TypeError** raised by trying to access the property `pc_` of **undefined** `this`.


VinceOPS:
Ok, nevermind, fixed by https://github.com/webrtc/apprtc/pull/490

Issue #384, opened 2016-10-10
collider install error: non-standard import "golang.org/x/net/websocket"


aprilsam:
**Browsers and versions affected**
CentOS7, go1.7.1

**Description**
I've installed go and write a Hello World to run, it's OK.
Then I entered 'go get collidermain' on terminal. It prompt me: cannot find package "golang.org/x/net/websocket" in any of:
	/program/go/src/vendor/golang.org/x/net/websocket (vendor tree)
	/program/go/src/golang.org/x/net/websocket (from $GOROOT)
	/program/go/work/src/golang.org/x/net/websocket (from $GOPATH)

So, I entered 'go get golang.org/x/net/websocket', 'go install golang.org/x/net/websocket' to get & install websocket. The package installed successfully.

But when I entered 'go get collidermain' on terminal, it still failed however. It print that:
/program/go/src/collider/collider.go:11:2: non-standard import "golang.org/x/net/websocket" in standard package "collider"

If anyone can help me solve this problem? Thanks!

**Steps to reproduce**
After installed go and linked the collider directories into $GOPATH/src
# go get golang.org/x/net/websocket
# go install golang.org/x/net/websocket
# go get collidermain

**Expected results**
No message appear.

**Actual results**
non-standard import "golang.org/x/net/websocket" in standard package "collider"

hustwei:
The same with me!

root@complie-desktop:/var/go/src# go get collider
can't load package: /var/go/src/collider/collider.go:11:2: non-standard import "golang.org/x/net/websocket" in standard package "collider"
root@complie-desktop:/var/go/src# go get collidermain
/var/go/src/collider/collider.go:11:2: non-standard import "golang.org/x/net/websocket" in standard package "collider"

KaptenJansson:
https://godoc.org/golang.org/x/net/websocket still says it's valid, I've not tried go 1.7.1. Could you try 1.6 and report back?

savayg:
got the same error and solved it fixing the workspace `GOPATH: /usr/local/goWorkspace` and properly linking collider files to the $GOPATH/src folder as explained in the docs.

KaptenJansson:
Aha I assumed everyone had followed the instructions ;)

aprilsam:
As KaptenJansson said,
> https://godoc.org/golang.org/x/net/websocket still says it's valid, I've not tried go 1.7.1. Could you try 1.6 and report back?

[...]

PR #494, opened 2017-10-19
Update README for clearer installation steps

nathando:
**Description**
It took me some time to figure out what is wrong when first trying to install this. And I've realized others have the same problem. It is because the README is unclear. 
For example: this issue is common: https://github.com/webrtc/apprtc/issues/384

**Purpose**
Make the README clearer so that less people will have issue installing

### DIFF
--- a/src/collider/README.md
+++ b/src/collider/README.md
@@ -4,8 +4,11 @@ A websocket-based signaling server in Go.
 
 ## Building
 
-1. Install the Go tools and workspaces as documented at http://golang.org/doc/install and http://golang.org/doc/code.html
-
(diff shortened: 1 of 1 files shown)


src/collider/README.md
```diff
@@ -23,10 +26,12 @@ A websocket-based signaling server in Go.
 
 5. Install dependencies
 
-        go get collidermain
+        go get golang.org/x/net/websocket
```
KaptenJansson: Are these really necessary since collidermain imports collider which has all the deps needed?

dyoh777: yes it is because 'go get collidermain' fails for so many people despite having working go environments and configurations per instructions

PR #517, opened 2017-12-08
remove callstats

KaptenJansson:
**Description**
Remove callstats

**Purpose**
No longer used.


chemmagate:
@KaptenJansson Hi, this PR breaks integrations for several users that use callstats with apprtc and also we use it for internal testing, is it possible to bring this integration back?

PR #519, opened 2017-12-15
Fix turn override for android and iOS

### LINKED ISSUE #456: How to use TURN_SERVER_OVERRIDE config in AppRTC-Android client?

KaptenJansson:
**Description**
- fixes #456 (uses the fix suggested in https://github.com/webrtc/apprtc/issues/456#issuecomment-351020935)
- Remove CEOD as it's no longer used.
- Update GCLOUD version
- Rename TURN_SERVER_OVERRIDE to ICE_SERVER_OVERRIDE
- Pass in the ICE_SERVER_OVERRIDE to pc config ice servers rather than using a separate attribute for this

**Purpose**
- Reduce the amount of changes required in the clients.

### DIFF
--- a/build/ensure_gcloud_sdk_is_installed.py
+++ b/build/ensure_gcloud_sdk_is_installed.py
@@ -14,7 +14,7 @@ import subprocess
 # https://cloud.google.com/sdk/downloads#versioned
 
 GCLOUD_DOWNLOAD_URL = 'https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/'
-GCLOUD_SDK_TAR_FILE = 'google-cloud-sdk-138.0.0-linux-x86_64.tar.gz'
+GCLOUD_SDK_TAR_FILE = 'google-cloud-sdk-183.0.0-linux-x86_64.tar.gz'
 GCLOUD_SDK_INSTALL_FOLDER = 'google-cloud-sdk'
 TEMP_DIR = 'temp'
 GCLOUD_SDK_PATH = os.path.join(TEMP_DIR, GCLOUD_SDK_INSTALL_FOLDER)
--- a/src/app_engine/apprtc.py
+++ b/src/app_engine/apprtc.py
@@ -42,12 +42,14 @@ def get_hd_default(user_agent):
   return 'true'
 
 # iceServers will be filled in by the TURN HTTP request.
-def make_pc_config(ice_transports):
+def make_pc_config(ice_transports, ice_server_override):
   config = {
   'iceServers': [],
   'bundlePolicy': 'max-bundle',
   'rtcpMuxPolicy': 'require'
   };
+  if ice_server_override:
+    config['iceServers'] = ice_server_override
   if ice_transports:
     config['iceTransports'] = ice_transports
   return config
@@ -260,7 +262,6 @@ def get_room_parameters(request, room_id, client_id, is_initiator):
   # TODO(tkchin): We want to provide a ICE request url on the initial get,
   # but we don't provide client_id until a join. For now just generate
   # a random id, but we should make this better.
-  # TODO(jansson): Remove this once CEOD is deprecated.
   username = client_id if client_id is not None else generate_random(9)
   if len(ice_server_base_url) > 0:
     ice_server_url = constants.ICE_SERVER_URL_TEMPLATE % \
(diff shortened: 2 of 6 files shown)


KaptenJansson:
I think we still need to modify the client to allow for turn override since it does not know that it should NOT connect to the ICE server.

coderxhs:
@KaptenJansson @MirkoBonadei i update the apprtc and webrtc  code to the latest version, it works perfect by Firefox on PC, but on android client(AppRTCMobile.apk), the problem is still there like below, can you give me some advise how to modify the webrtc to adapt the apprtc? the app is still try to connect the ICE server,but i used a coturn(https://github.com/coturn/coturn).
          
unnamed

didactics:
Hi @coderxhs! As @Pizza67 pointed out (https://github.com/webrtc/apprtc/issues/456#issuecomment-351020935), the [Android AppRTC](https://webrtc.googlesource.com/src/+/master/examples/androidapp/src/org/appspot/apprtc/RoomParametersFetcher.java) needs to be modified to accept the TURN/STUN overrides.

To configure my local instance of [AppRTC server](https://github.com/webrtc/apprtc.git) for external TURN/STUN servers, in [constants.py](https://github.com/webrtc/apprtc/blob/master/src/app_engine/constants.py) I commented out [this](https://github.com/webrtc/apprtc/blob/master/src/app_engine/constants.py#L23) and uncommented [ICE_SERVER_OVERRIDE](https://github.com/webrtc/apprtc/blob/master/src/app_engine/constants.py#L25) according to @KaptenJansson's recommendations. (Heads up, after uncommenting the [ICE_SERVER_OVERRIDE](https://github.com/webrtc/apprtc/blob/master/src/app_engine/constants.py#L25) line, remove the leading indent, or the built and running server won't accept connections.) In my case the config looked like this:

```
ICE_SERVER_OVERRIDE  = [
 {
   "urls": [
     "turn:13.250.13.83:3478?transport=udp",
     "turn:13.250.13.83:3478?transport=tcp"
   ],
   "username": "YzYNCouZM1mhqhmseWk6",
   "credential": "YzYNCouZM1mhqhmseWk6"
 },
 {
   "urls": [
     "stun:hostname/IpToStunServer:19302"
   ]
 }
]
```
Then I had to modify the [Android client app](https://webrtc.googlesource.com/src/+/master/examples/androidapp/src/org/appspot/apprtc/RoomParametersFetcher.java#200). Lines 200 through 218 should be replaced with:
```
// Return the list of ICE servers described by a WebRTCPeerConnection
// configuration string.
private List<PeerConnection.IceServer> iceServersFromPCConfigJSON(String pcConfig)
    throws JSONException {
  JSONObject json = new JSONObject(pcConfig);
  JSONArray servers = json.getJSONArray("iceServers");
  List<PeerConnection.IceServer> ret = new ArrayList<>();
  for (int i = 0; i < servers.length(); ++i) {
    JSONObject server = servers.getJSONObject(i);
    JSONArray turnUrls = server.getJSONArray("urls");
    String username = server.has("username") ? server.getString("username") : "";
    String credential = server.has("credential") ? server.getString("credential") : "";

    for (int j = 0; j < turnUrls.length(); j++) {
      String turnUrl = turnUrls.getString(j);
      PeerConnection.IceServer turnServer =
          PeerConnection.IceServer.builder(turnUrl)
                  .setUsername(username)
                  .setPassword(credential)
                  .createIceServer();
      ret.add(turnServer);
    }
  }
  return ret;
}
```
With these changes, both Firefox and the Android client app join in a room successfully. If the changes to the client are accepted, hope to see them in the Google repo one day.

PR #597, opened 2018-12-26
Using Forwarded header for scheme if FORWARD_SCHEME env variable is True.

charlie-mtz:
**Description**
Added logic to replace request.scheme with Forwarded HTTP header ([RFC 7239](https://tools.ietf.org/html/rfc7239)) if environment variable FORWARD_SCHEME=True.

**Purpose**
Allows using Apprtc behind a TLS termination proxy. Makes Apprtc easier to deploy in a Docker environment.

**Test**
Example usage in the test development server:
`dev_appserver.py apprtc/out/app_engine --env_var FORWARD_SCHEME=True`

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,6 +1,7 @@
 // Karma configuration
 module.exports = function(config) {
-  var browser = process.env.BROWSER || 'chrome';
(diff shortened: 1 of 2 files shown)


KaptenJansson:
Thanks for doing this, could you also add instructions to the README under the DEPLOYMENT section?

PR #600, opened 2019-01-07
Replace insecure Grunt-htmlhint with grunt-htmllint but leave it disabled due to 17 big blocking warnings

KaptenJansson:
**Description**
Remove insecure deps, add Python pip install instructions and commit a new package-lock.json.

**Purpose**
Prepare for other changes.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -34,14 +34,13 @@ module.exports = function(grunt) {
       }
     },
 
-    htmlhint: {
-      html1: {
+    htmllint: {
+      all: {
         src: [
           'src/**/*_template.html'
         ]
       }
     },
-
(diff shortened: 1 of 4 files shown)


Gruntfile.js
```diff
@@ -178,7 +177,7 @@ module.exports = function(grunt) {
   // Set default tasks to run when grunt is called without parameters.
   grunt.registerTask('default', ['runLinting', 'runPythonTests', 'build',
                                  'runUnitTests']);
-  grunt.registerTask('runLinting', ['csslint', 'htmlhint', 'eslint']);
```
KaptenJansson: Leaving out htmllint here on purpose as mentioned in the summary.

PR #601, opened 2019-01-07
Set async call cleanup as sync XHR calls are banned in Chrome

KaptenJansson:
**Description**
Set async call cleanup as sync XHR calls are banned in Chrome

**Purpose**
Comply with web standards. If this works in downstream projects, we can remove the sync/async logic and make it async permanently.

Fixes browserstests in https://bugs.chromium.org/p/chromium/issues/detail?id=918871 hopefully.

Base is set to the updateDeps branch.

### DIFF
--- a/src/web_app/js/appcontroller.js
+++ b/src/web_app/js/appcontroller.js
@@ -220,7 +220,7 @@ AppController.prototype.finishCallSetup_ = function(roomId) {
     // clean up steps before page is closed.
     // Chrome apps can't use onbeforeunload.
     window.onbeforeunload = function() {
-      this.call_.hangup(false);
+      this.call_.hangup(true);
     }.bind(this);
 
     window.onpopstate = function(event) {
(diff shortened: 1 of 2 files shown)


oprypin:
This fix makes the test from crbug/918871 pass. Thanks.

oprypin:
But it's possible that these XHRs won't actually be fulfilled, and that only the `fetch` API has the ability to finish a request after page close.

KaptenJansson:
They are currently wrapped in a promise and I guess promises survive the page close? Or are does only white listed API's that happens to use Promises (promises themselves do not survive?) do this?

KaptenJansson:
Fetch is essentially XHR wrapped in a promise by the looks of it.

KaptenJansson:
Ok, will keep an eye on it.

oprypin:
> Merged #601 into **updateDeps**.

Probably needs to be merged into master instead??

Issue #552, opened 2018-03-29
import requests Module Not Found


adenkaplan:
**Description**
When we run "grunt build" we got an error

**Steps to reproduce**
step 1 - npm install
step 2 - install python 3.6.5
step 3 - grunt build

**Actual results**
Warning: Command failed: python ./build/build_app_engine_package.py src out/app_engine
Traceback (most recent call last):
  File "./build/build_app_engine_package.py", line 12, in <module>
    import requests
ModuleNotFoundError: No module named 'requests'
 Use --force to continue.

Aborted due to warnings.

nagarjunbn:
I had the same error . 
For ubuntu use below commands to install requests .
`apt-get install python-requests`
`sudo pip install requests` or `sudo pip3 install requests`

PR #604, opened 2019-01-18
Update README with comment about pip requirements

kuddai:
**Description**
Project has additional python dependencies. It would be good to mention them in the installation/deployment section. 
**Purpose**
Project already has explicitly stated  python dependencies in requirements.txt. By mentioning this installation step in README.md it might help avoid issues like [this](https://github.com/webrtc/apprtc/issues/552).


kuddai:
I signed it!

PR #638, opened 2019-09-19
Add setOpusOptions to setLocalSDP

### LINKED ISSUE #23: AppRTC: URL param stereo=true does not result in stereo due to APM still enabled
https://github.com/GoogleChrome/webrtc/issues/220#issuecomment-72776788

KaptenJansson commented on Nov 17, 2014
The URL param Stereo=true does not enable stereo all the way (just sets the stereo=1 in the SDP) in chrome unless echoCancellation=false is sent in as an optional gUM constraint.

This is due to the APM does not handle stereo hence it's down-mixed to mono. 
The URL params &stereo=true, audio=echoCancellation=false will result in stereo.

If the intention of stereo=true is to enable proper stereo (not just setting the SDP attribute) it should also pass the audio=echoCancellation=false gum constraint.

juberti commented 15 days ago
Is AEC the only thing that doesn't tolerate stereo? Or does all APM processing have to be disabled? @KaptenJansson can you check with Tina?
Christoffer Jansson
 Owner
 KaptenJansson commented 14 days ago
I believe the APM converts to mono before processing the audio hence it
affects all processing. But I will double check with Tina.
…

KaptenJansson:
**Description**
Sets audio options in the sdp.

**Purpose**
Fixes #23

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,8 +30,8 @@ matrix:
     - env: BROWSER=chrome  BVER=unstable
 
 before_install:
-  - nvm install 6
-  - nvm use 6
+  - nvm install 8
+  - nvm use 8
 
 before_script:
   - npm install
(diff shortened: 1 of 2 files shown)


KaptenJansson:
Eslint works now but there are so many existing errors it's not viable to fix in this PR, will do in a separate PR. Will merge anway.

PR #657, opened 2020-05-05
fix sdes loopback more

fippo:
until #312 gets done this works around
  https://bugs.chromium.org/p/chromium/issues/detail?id=1077740

### DIFF
--- a/src/web_app/js/loopback.js
+++ b/src/web_app/js/loopback.js
@@ -70,8 +70,9 @@ function setupLoopback(wssUrl, roomId) {
       // this line but for loopback where we reuse the offer, that is skipped
       // and remains in the answer and breaks the call.
       // https://bugs.chromium.org/p/chromium/issues/detail?id=616263
+      // https://bugs.chromium.org/p/chromium/issues/detail?id=1077740
       loopbackAnswer = loopbackAnswer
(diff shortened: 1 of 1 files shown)


src/web_app/js/loopback.js
```diff
@@ -70,8 +70,10 @@ function setupLoopback(wssUrl, roomId) {
       // this line but for loopback where we reuse the offer, that is skipped
       // and remains in the answer and breaks the call.
       // https://bugs.chromium.org/p/chromium/issues/detail?id=616263
+      // https://bugs.chromium.org/p/chromium/issues/detail?id=1077740
       loopbackAnswer = loopbackAnswer
-          .replace(/a=crypto:0 AES_CM_128_HMAC_SHA1_32\sinline:.{44}/, '');
+          .replace(/a=crypto:1 .*/, '')
+          .replace(/a=crypto:2 .*/, '');
```
alvestrand, line 76: should this be .replace(/a=crypto:[1-9]+ .*/g, '')?
are we guaranteed that the first crypto algo is always OK to use?

fippo: I *think* we are but this is sdes and it seems like chrome might even not be compliant in not handling multiple a=crypto in an answer.
Updated

fippo:
lint failure is... meh.

PR #666, opened 2020-06-17
Update loopback.js

jackzelenkin:
**Description**
On a Mac, loopbackAnswer contains \\r\\n characters instead of proper line breaks. Because of that initial regex /a=crypto:[1-9]+ .*/g matches everything starting from a=crypto:1 and all the way to the end of loopbackAnswer, and when replaced for an empty string it corrupts the json. 

**Solution**
As a solution, I propose to first search-and-replace "mac-specific" crypto patterns (crypto all the way to first \\r\\n sequence) and only after that search-and-replace original patterns (crypto all the way to the end of the line).

### DIFF
--- a/src/web_app/js/loopback.js
+++ b/src/web_app/js/loopback.js
@@ -72,6 +72,7 @@ function setupLoopback(wssUrl, roomId) {
       // https://bugs.chromium.org/p/chromium/issues/detail?id=616263
       // https://bugs.chromium.org/p/chromium/issues/detail?id=1077740
       loopbackAnswer = loopbackAnswer
+          .replace(/a=crypto:[1-9]+ .*?\\r\\n/g, '')
           .replace(/a=crypto:[1-9]+ .*/g, '');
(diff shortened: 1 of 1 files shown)


jackzelenkin:
@googlebot I signed it!

alvestrand:
Wrong fix, I think.
\r\n is the specified line termination character in SDP, \n is exceptionally (but universally) accepted.
@fippo should look at this - either the pattern leaves a blank line (just the \r\n or \n), or it eats the \n when there's no \r and doesn't eat it when it's present.
I think you need extra regex modifiers to get substitutes to work past linebreaks (of any type).

fippo:
The code takes loopbackAnswer as wssMessage.msg which is the serialized JSON which escapes all the SDP crlfs as \\r\\n. That is a bit silly and unexpected...
And yes, the pattern leaves blank lines, thank you! #667 should fix both

jackzelenkin:
Thanks Philipp!

PR #677, opened 2020-09-07
Stop building the chrome app and cleanup unmaintained dependencies.

kisabaka:
This allows building the app and running the JS unit tests with less hassle. Before this change, it was easy to get in a state where the closure compiler wouldn't build anymore after updating node modules.

**Description**
Remove chrome app, stop using `grunt-closurecompiler-new-grunt` and fix unit tests.

**Purpose**
Cleanup unmaintained dependencies.


KaptenJansson (approved):
LGTM.

FIY, the closure compilation was done in order to speed up page load and reduce the page footprint. This is much less of an issue today and also I actually prefer a non compiled version of this as it's a Demo app, not a production application thus having non compiled code will make it easier to understand.

PR #679, opened 2020-09-09
Add a way to pass the TURN server API key as a request parameter.

kisabaka:
**Description**

**Purpose**
This allows using a different key than the one configured in app.yaml
through the new 'apikey=' parameter, just like we do with the TURN server
using the 'ts' parameter.


KaptenJansson (approved):
LGTM

PR #685, opened 2020-09-14
Show a persistent warning about TURN through the call

kisabaka:
**Description**
Show a message warning the user that the call might not work over the network when the request to get TURN servers fails.

**Purpose**


MirkoBonadei (approved):
LGTM

PR #708, opened 2021-06-22
Fix wrong guide information from apprtc params

HuaChunbo:
**Description**
googNoiseSuppression is the correct one for Audio NS.
googNoiseReduction is for video frame denoise.

**Purpose**
This patch tells the right information about media constraints usage of "audio=googNoiseSuppression=false" or "video=googNoiseReduction=false".


fippo:
there seems to be a considerable amount of confusion here:
https://source.chromium.org/search?q=googNoiseReduction&sq=
https://source.chromium.org/search?q=googNoiseSuppression&sq=

@guidou can you try to find someone who understands the difference?

guidou:
cc @alebzk

guidou:
Checked internally with audio people and this change looks good.

guidou (approved):

HuaChunbo:
Thanks for the quick merge. Hope the live site of params.html can be updated soon as well, so that APPRTC users would not be misled any more.

fippo:
@KaptenJansson can you deploy please?

PR #720, opened 2021-12-16
Add Dockerfile for localhost non TLS docker image

KaptenJansson:
**Description**
Sets up a localhost non TLS capable AppRTC instance with Collider

I've made it the way it is to keep everything contained inside the Dockerfile without requiring other scripts etc (hence why the start.sh script is generated within the Dockerfile).

**Purpose**
Allow local development and testing using AppRTC locally

NOTE: Any docker specialist please chime in, I'm quite the noob in dockerverse

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,8 +57,8 @@ CMD ./start.sh
 # - Build the Dockerfile into an image: 'sudo docker build apprtc/'
 #   Note the image ID from the build command, e.g. something like 'Successfully built 503621f4f7bd'.
 # - Run: 'sudo docker run -p 8080:8080 -p 8089:8089 --rm -ti 503621f4f7bd'
-#   The container will now run in interactive mode and output logging. This can ofc be prevent if you run in 
-#   non-interactive mode. The -p are port mappings to the GAE app and Collider instances, the host ones can be changed.
+#   The container will now run in interactive mode and output logging. If you do not want this, omit the '-ti' argument.
+#   The '-p' options are port mappings to the GAE app and Collider instances, the host ones can be changed.
 #
 # - On the same machine that this docker image is running on you can now join apprtc calls using 
 #   http://localhost:8080/?wshpp=localhost:8089&wstls=false,  once you join the URL will have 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+
+# Start the bash wrapper that keeps both collider and the AppRTC GAE app running. 
+CMD ./start.sh
+
+## Instructions (Tested on Debian 11 only):
+# - Download the Dockerfile from the AppRTC repo and put it in a folder, e.g. 'apprtc'
+# - Build the Dockerfile into an image: 'sudo docker build apprtc/'
+#   Note the image ID from the build command, e.g. something like 'Successfully built 503621f4f7bd'.
+# - Run: 'sudo docker run -p 8080:8080 -p 8089:8089 --rm -ti 503621f4f7bd'
+#   The container will now run in interactive mode and output logging. This can ofc be prevent if you run in
```
alvestrand, line 60: Nit: "ofc be prevent" -> "of course be prevented"
Should you tell how to run a docker in non-interactive mode? For me as a non-user of Docker, it's not obvious.

MirkoBonadei (approved):

Dockerfile
KaptenJansson: Done

eakraly:
Maybe a cosmetic comment: this file creates an image with 26 layers
This by itself not an issue but I would merge some of the layers - for example downloading/unpacking/running google-cloud-sdk are 3 different layers. This will not help with caching much if a version changes (all 3 will be rebuilt)

KaptenJansson:
Thank you for taking time reviewing!

> Maybe a cosmetic comment: this file creates an image with 26 layers This by itself not an issue but I would merge some of the layers - for example downloading/unpacking/running google-cloud-sdk are 3 different layers. This will not help with caching much if a version changes (all 3 will be rebuilt)

Roger, good to know and was wondering about this when I saw the amount of steps when building.

> 
> Or creating a start.sh using 4 layers...
> 
> Additional benefit of having a "full flow" of installing something in a single RUN command is that the downloaded zip and unpacked folder can be deleted in the same layer so that the end result image is smaller. (I get an image of 1.8GB of which unzipped gcloud.tar.gz is 459MB - which is not used after being installed)
> 
> Again, not a big deal if this is only to be used for local development.

Done! I've reduced the number of steps in general.

PR #723, opened 2021-12-31
docker: Fix stunnel download URL

philn:
The /downloads/ path seems to host only the most recent version of the library,
so fallback to the /archive/ path, which has all the versions.


KaptenJansson:
Nice find, thanks for fixing!

KaptenJansson (approved):

PR #727, opened 2022-02-09
Fix docker run option about port

SpringMT:
**Description**
To access this URL `https://localhost:8080/?wshpp=localhost:8089&wstls=true`, need to change the docker run option `docker run -p 443:443` to `docker run -p 8080:443`.

**Purpose**


SpringMT:
Fixed https://github.com/webrtc/apprtc/commit/0430747bd0a4bd3aa08fd9c42231773b776241c2

PR #756, opened 2023-03-10
Added arguments to make the server accessible remotely.

bkeller-vuzix:
The original flags were requiring TLS be used, and that remote machines be whitelisted. Those are now docker build arguments so they can be disabled.

Also changed the instructions to tag the docker image so the command is more consistent.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,16 @@
 
 FROM golang:1.17.5-alpine3.15
 
+# Build arguments.
+#
+# Override on the docker command line with
+#   --build-arg USE_TLS_SOCKETS=false
+#   --build-arg ENABLE_HOST_CHECK=false
+#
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -2,6 +2,16 @@
 
 FROM golang:1.17.5-alpine3.15
 
+# Build arguments.
+#
+# Override on the docker command line with
+#   --build-arg USE_TSL_SOCKETS=false
+#   --build-arg ENABLE_HOST_CHECK=false
+#
+ARG USE_TSL_SOCKETS=true
```
alvestrand: Shouldn't this be TLS_SOCKETS?

bkeller-vuzix: Thank you. You are definitely correct. I have updated that commit to remove that typo.

